PR #1, opened 2015-09-10
Update README.md to provide an introduction

mrisher:
Maybe I can interview you and help further revise the intro. I think we still need more of a "why would I want this" and "how does it compare to the competition" to encourage adoption.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,13 @@
 # BadWolf
 
-BadWolf was born as a loosely modeled graph store after RDF. However, triples
-were expanded to quads to allow simpler temporal reasoning. Most of the web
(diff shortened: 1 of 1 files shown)


xllora:
Yes, that would be awesome if you have the time and will!

Working on that is on the bottom of my todo's. I am currently focusing on having this new version of the language working end to end for all the statements in BQL. I have now the create, drop, insert and delete statements working e2e. I am half way on the work on the semantic for the for the select (specifically the graph clauses). 

Also, remember this is just the abstraction layer and only has an in memory volatile back end. (I hope some of our friends get excited and pick up writing the driver for https://cloud.google.com/bigtable/ :)) so there is still a bit of road ahead for how does it compare to other options ;)

PR #35, opened 2016-02-18
Implement todos

defrager:
Unquote string correctly.
Don't allow partial substrings prefix for covariant (or maybe it should be ok?)


triple/node/node.go
```diff
@@ -36,8 +36,11 @@ func (t *Type) String() string {
 // Covariant checks for given two types A and B, A covariant B if B _is a_ A.
 // In other word, A _covariant_ B if B is a prefix of A.
 func (t *Type) Covariant(ot *Type) bool {
-	// TODO: is /ab covariant of /a ?
-	return strings.HasPrefix(t.String(), ot.String())
+	if !strings.HasPrefix(t.String(), ot.String()) {
+		return false
+	}
+	// /type/foo is covarian of /type, but /typefoo is not covatiant of /type.
+	return len(t.String()) == len(ot.String()) || t.String()[len(ot.String())] == '/'
```
xllora: Missing -1 at?

return len(t.String()) == len(ot.String()) || t.String()[len(ot.String())-1] == '/'

triple/predicate/predicate.go
```diff
@@ -80,7 +81,11 @@ func Parse(s string) (*Predicate, error) {
 	if idx < 0 {
 		return nil, fmt.Errorf("predicate.Parse could not find anchor definition in %s", raw)
 	}
-	id, ta := raw[1:idx], raw[idx+3:len(raw)-1]
+	id, ta := raw[0:idx+1], raw[idx+3:len(raw)-1]
```
xllora: Actually I have been using them as optional. The documentation is a bit outdate.

xllora:
Quick question for you.

triple/node/node.go
defrager: nope.
if B is "/type"
and A is "/type/foo"
we expect "/" in A[5] index.

xllora: Do'h. Right. I was reading it the other way around. Thx!

PR #44, opened 2016-06-02
Fix typos and bw build instructions

cdusovic:


xllora:
Thanks Christopher for all the proofreading 👍

PR #64, opened 2017-03-03
Issue 46: Parse Construct Statement

palimarrao:
Also fixed some typos that I found.

### DIFF
--- a/bql/grammar/grammar.go
+++ b/bql/grammar/grammar.go
@@ -204,15 +204,6 @@ func BQL() *Grammar {
 				},
 			},
 		},
-		"CONSTRUCT_FACTS": []*Clause{
-			{
-				Elements: []Element{
-					NewTokenType(lexer.ItemLBracket),
-					NewSymbol("CLAUSES"),
-					NewTokenType(lexer.ItemRBracket),
-				},
-			},
-		},
 		"CLAUSES": []*Clause{
 			{
 				Elements: []Element{
@@ -751,6 +742,94 @@ func BQL() *Grammar {
 					NewTokenType(lexer.ItemPredicate),
 					NewSymbol("DELETE_OBJECT"),
 					NewSymbol("DELETE_DATA"),
+				},
+			},
+			{},
+		},
+		"CONSTRUCT_FACTS": []*Clause{
+			{
+				Elements: []Element{
+					NewTokenType(lexer.ItemLBracket),
+					NewSymbol("CONSTRUCT_TRIPLES"),
+					NewTokenType(lexer.ItemRBracket),
+				},
+			},
+		},
+		"CONSTRUCT_TRIPLES": []*Clause{
+			{
+				Elements: []Element{
+					NewTokenType(lexer.ItemNode),
+					NewSymbol("CONSTRUCT_PREDICATE"),
+					NewSymbol("CONSTRUCT_OBJECT"),
+					NewSymbol("MORE_CONSTRUCT_TRIPLES"),
+				},
+			},
+			{
+				Elements: []Element{
+					NewTokenType(lexer.ItemBlankNode),
+					NewSymbol("CONSTRUCT_PREDICATE"),
+					NewSymbol("CONSTRUCT_OBJECT"),
+					NewSymbol("MORE_CONSTRUCT_TRIPLES"),
+				},
+			},
+			{
+				Elements: []Element{
+					NewTokenType(lexer.ItemBinding),
+					NewSymbol("CONSTRUCT_PREDICATE"),
+					NewSymbol("CONSTRUCT_OBJECT"),
+					NewSymbol("MORE_CONSTRUCT_TRIPLES"),
+				},
+			},
+		},
+		"CONSTRUCT_PREDICATE": []*Clause{
+			{
+				Elements: []Element{
+					NewTokenType(lexer.ItemPredicate),
+				},
+			},
+			{
+				Elements: []Element{
+					NewTokenType(lexer.ItemPredicateBound),
+				},
+			},
+			{
+				Elements: []Element{
+					NewTokenType(lexer.ItemBinding),
+				},
+			},
+		},
+		"CONSTRUCT_OBJECT": []*Clause{
+			{
+				Elements: []Element{
+					NewTokenType(lexer.ItemNode),
+				},
+			},
+			{
+				Elements: []Element{
+					NewTokenType(lexer.ItemPredicate),
+				},
+			},
+			{
+				Elements: []Element{
+					NewTokenType(lexer.ItemPredicateBound),
+				},
+			},
+			{
+				Elements: []Element{
+					NewTokenType(lexer.ItemLiteral),
+				},
+			},
+			{
+				Elements: []Element{
+					NewTokenType(lexer.ItemBinding),
+				},
+			},
+		},
+		"MORE_CONSTRUCT_TRIPLES": []*Clause{
+			{
+				Elements: []Element{
+					NewTokenType(lexer.ItemDot),
+					NewSymbol("CONSTRUCT_TRIPLES"),
 				},
 			},
 			{},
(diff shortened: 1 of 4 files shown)


palimarrao:
The latest commit handles triples and blank nodes. The reification ; syntax will be implemented in a different commit.

bql/grammar/grammar.go
```diff
@@ -733,6 +746,94 @@ func BQL() *Grammar {
...
+		},
+		"CONSTRUCT_PREDICATE": []*Clause{
+			{
+				Elements: []Element{
+					NewTokenType(lexer.ItemPredicate),
+				},
+			},
+			{
+				Elements: []Element{
+					NewTokenType(lexer.ItemPredicateBound),
```
xllora: Predicate bounds can present a number of predicates. I am not sure we want this here. I think we should only support predicate and binding. For instance, if we go with

construct{
  ?s ?p@[,some_ts] ?o
}

Does it mean we have to filter further the binded predicates of ?p. This seems a bit redundant since it can be done by construing the where clause further or by adding a having clause. For that reason, I would suggest dropping it.

bql/grammar/grammar.go
```diff
@@ -733,6 +746,94 @@ func BQL() *Grammar {
...
+				},
+			},
+			{
+				Elements: []Element{
+					NewTokenType(lexer.ItemPredicate),
+				},
+			},
+			{
+				Elements: []Element{
+					NewTokenType(lexer.ItemPredicateBound),
```
xllora: Same comment here about the predicate bound.

bql/grammar/grammar_test.go
```diff
@@ -113,6 +113,16 @@ func TestAcceptByParse(t *testing.T) {
 		                          /room<000> "connects_to"@[] /room<001>};`,
 		`delete data from ?world {/room<000> "named"@[] "Hallway"^^type:text.
 		                          /room<000> "connects_to"@[] /room<001>};`,
+		// Test Construct clause.
+		`construct {?s "foo"@[,] ?o} into ?a from ?b where {?s "foo"@[,] ?o} having ?s = ?o;`,
```
xllora: This is exactly the point I was making above. If you want to create triples for each of the `foo` predicates with the right timestamp, you could write it like this instead

```
CONSTRUCT {?s ?p ?s} 
INTO ?a 
FROM ?b 
WHERE {
  ?s "foo"@[,] as ?p ?s
} ;
```

Also see how you can also avoid the having by properly playing with the binding, but that is beside the point here, since we should definitely have test that cover the having clause :smile:

bql/grammar/grammar_test.go
```diff
@@ -344,7 +365,7 @@ func TestRejectByParseAndSemantic(t *testing.T) {
 	}
 }
 
-func TestSemanticStatementGraphClausesLenghtCorrectness(t *testing.T) {
+func TestSemanticStatementGraphClausesLengthCorrectness(t *testing.T) {
```
xllora: Thx for the typo fixing :smile:

bql/lexer/lexer.go
```diff
@@ -603,6 +616,22 @@ func lexNode(l *lexer) stateFn {
...
 }
 
+// lexBlankNode tries to lex a blank node out of the input
+func lexBlankNode(l *lexer) stateFn {
+	if r := l.next(); r != colon {
+		l.emitError("blank node should start with _:")
+		return nil
+	}
+	for {
+		if r := l.next(); !unicode.IsLetter(r) && !unicode.IsDigit(r) && r != rune('_') || r == eof {
```
xllora: Do we want to force the first rune to be a letter? right now this `_:_` or `_:1` would be valid blank node IDs?

bql/lexer/lexer.go
```diff
@@ -618,7 +647,7 @@ func lexPredicateOrLiteral(l *lexer) stateFn {
 	return lexLiteral
 }
 
-// lexPredicate lexes a predicate of out of the input.
```
xllora: Thx :smile:

bql/lexer/lexer.go
```diff
@@ -669,7 +698,7 @@ func lexPredicate(l *lexer) stateFn {
 	return lexSpace
 }
 
-// lexPredicate lexes a literal of out of the input.
+// lexLiteral lexes a literal out of the input.
```
xllora: yay! :smile_cat:

bql/lexer/lexer_test.go
```diff
@@ -101,6 +102,16 @@ func TestIndividualTokens(t *testing.T) {
 				{Type: ItemError, Text: "/_<foo",
 					ErrorMessage: "[lexer:0:6] node is not properly terminated; missing final > delimiter"},
 				{Type: ItemEOF}}},
+		{"_:v1 _:foo_bar",
+			[]Token{
+				{Type: ItemBlankNode, Text: "_:v1"},
+				{Type: ItemBlankNode, Text: "_:foo_bar"},
+				{Type: ItemEOF}}},
+		{"_v1",
```
xllora: See my comment about if we should start with a letter always the ID of the blank node.

bql/lexer/lexer.go
palimarrao: That's a good idea. The BNode label name now should begin with a letter.

bql/grammar/grammar.go
palimarrao: Got it. I have removed the support for binded predicates.

bql/grammar/grammar.go
palimarrao: Done.

bql/grammar/grammar_test.go
palimarrao: I have updated this test case.

palimarrao:
Thanks for the detailed review @xllora ! I have addressed your comments.

bql/grammar/grammar_test.go
```diff
@@ -113,6 +113,16 @@ func TestAcceptByParse(t *testing.T) {
 		                          /room<000> "connects_to"@[] /room<001>};`,
 		`delete data from ?world {/room<000> "named"@[] "Hallway"^^type:text.
 		                          /room<000> "connects_to"@[] /room<001>};`,
+		// Test Construct clause.
+		`construct {?s "foo"@[] ?o} into ?a from ?b where {?s "foo"@[,] ?o} having ?s = ?o;`,
+		`construct {?s "foo"@[] ?o} into ?a from ?b where {?s "foo"@[,] ?o};`,
```
xllora: Could you use a different predicate other than "foo" to avoid confusing. Above it adds a foo that is immutable as opposed to the temporal ones you are querying.

On another note we may want to provide some way to allow folks to construct temporal predicates. Immutable predicates are easy, as you shown in the above examples. But for temporal ones, we would need something along the lines of:

TEMPORAL_PREDICATE("bar"^^type:string, ?ts)
TEMPORAL_PREDICATE("bar"^^type:string, @1234)
TEMPORAL_PREDICATE("bar"^^type:string, "2017....")

What do you thing? No need to do it in this round, but maybe we should file an issue for later.

bql/grammar/grammar_test.go
```diff
@@ -113,6 +113,16 @@ func TestAcceptByParse(t *testing.T) {
...
+		`construct {?s ?p ?o} into ?a from ?b where {?s "foo"@[,] ?o} having ?s = ?o;`,
```
xllora: Nit pick, since it will mater later, ?p is not defined and hence this statement is not going to be semantically valid. Other statements have a similar issue :)

palimarrao: Done.

bql/grammar/grammar_test.go
palimarrao: Done! For temporal predicates, can we use the same syntax as is used in the INSERT statement? 
`construct {?s "some_pred"@[2006-01-02T15:04:05.999999999Z07:00] ?o } into ?a from ?b where {?s "foo"@[,] ?o};`
Or is the reason that we might need to construct temporal predicates from predicates returned by the WHERE clause, but with a different time anchor?

palimarrao:
Thanks for the review @xllora !

PR #66, opened 2017-05-17
Added ConstructClause to Statement struct

palimarrao:
I want to make sure that the addition I am making to the `Statement` struct looks good.

### DIFF
--- a/bql/semantic/semantic.go
+++ b/bql/semantic/semantic.go
@@ -47,6 +47,8 @@ const (
 	Create
 	// Drop statement.
 	Drop
+	// Construct statement.
+	Construct
 )
 
 // String provides a readable version of the StatementType.
@@ -62,6 +64,8 @@ func (t StatementType) String() string {
 		return "CREATE"
 	case Drop:
 		return "DROP"
+	case Construct:
+		return "CONSTRUCT"
 	default:
 		return "UNKNOWN"
 	}
@@ -75,6 +79,8 @@ type Statement struct {
 	data                      []*triple.Triple
 	pattern                   []*GraphClause
 	workingClause             *GraphClause
+	constructClauses          []*ConstructClause
+	workingConstructClause    *ConstructClause
 	projection                []*Projection
 	workingProjection         *Projection
 	groupBy                   []string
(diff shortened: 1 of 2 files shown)


bql/semantic/semantic.go
```diff
@@ -75,6 +79,8 @@ type Statement struct {
 	data                      []*triple.Triple
 	pattern                   []*GraphClause
 	workingClause             *GraphClause
+	constructClauses          []*ConstructClause
```
xllora: Could you elaborate why you need a new working clause? The where clause of a construct is the  same as the one on a regular select. I was wondering if it is just cleaner to just add the actual construct information containing the info to create the new triples. Or am I just mislead by the name?

bql/semantic/semantic.go
```diff
@@ -122,6 +128,33 @@ type GraphClause struct {
 	OTemporal        bool
 }
 
+// ConstructClause represents a singular clause within a construct statement.
+type ConstructClause struct {
```
xllora: Despite the name this is just the info on the construct right? Just want to make sure I got it right :)

palimarrao: Yep, it is just the information required to insert a new triple into the graph. For instance (a contrived example):

`construct {?s ?p ?o; ?p1 ?o1} into ?a from ?b where {?s ?p ?o. ?s ?p1 ?o2}`

will lead to the creation of the following instance of `constructClauses`:

```
[]*ConstructClauses{
  {
    SBinding: ?s,
    PBinding: ?p,
    OBinding: ?o,
    ReificationClauses: []*ReificationClause{
      {
         PBinding: ?p1
         OBinding: ?o1
      },
   },
}

```
I guess the name is confusing. Do you have any suggestions?

bql/semantic/semantic.go
palimarrao: So `constructClauses` captures only the declarations within a `CONSTRUCT` statement that specifies how the triples must be built. For instance, consider the following:

`construct {?s "new_predicate"@[] ?o} into ?a from ?b where {?s "old_predicate"@[] ?o}`

`constructClauses` simply captures the `?s "new_predicate"@[] ?o` part. `workingClauses` will capture the `?s "old_predicate"@[] ?o` as before. `constructClauses` will capture multiple such declarations (separated by the `.` operator) and `constructClause` also has support for multiple reification clauses. So in a way, `constructClause` is similar to `data` within the `Statement` struct (except it has to support literals, bindings and reification). Does this seem like the right thing to do?

PR #68, opened 2017-05-23
Modifed Parse to handle Blank Nodes.

palimarrao:
I did not do too much of error checking in the Parse method since the lexer should check for valid blank nodes.

### DIFF
--- a/triple/node/node.go
+++ b/triple/node/node.go
@@ -101,9 +101,12 @@ func Parse(s string) (*Node, error) {
 		}
 		return NewNode(t, id), nil
 	case underscore:
-		id := ID(raw[2:len(raw)])
-		t :=  Type("/_")
-		return NewNode(&t, &id), nil
+		id, err := NewID(raw[2:len(raw)])
+		if err != nil {
+			return nil, fmt.Errorf("node.Parser: invalid ID in %q, %v", raw, err)
+		}
+		t, _ :=  NewType("/_")
+		return NewNode(t, id), nil
 	default:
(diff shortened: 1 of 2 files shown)


triple/node/node.go
```diff
@@ -74,25 +79,34 @@ func (n *Node) String() string {
...
+		if raw[len(raw)-1] != '>' {
+			return nil, fmt.Errorf("node.Parser: pretty printing should finish with '>' in %q", raw)
+		}
+		id, err := NewID(raw[idx+1 : len(raw)-1])
+		if err != nil {
+			return nil, fmt.Errorf("node.Parser: invalid ID in %q, %v", raw, err)
+		}
+		return NewNode(t, id), nil
+	case underscore:
+		id := ID(raw[2:len(raw)])
```
xllora: id := NewID(raw[2:len(raw)]0

triple/node/node.go
```diff
@@ -74,25 +79,34 @@ func (n *Node) String() string {
...
+		t :=  Type("/_")
+		return NewNode(&t, &id), nil
```
xllora: return NewNode(&t, id), nil

Both changes for consistency with the code below. I would also suggest use NewType.

triple/node/node.go
palimarrao: Done.

triple/node/node.go
palimarrao: Done.

palimarrao:
I have addressed the comments. PTAL, thanks!

PR #74, opened 2017-06-14
Added input and output graph lists.

palimarrao:
This PR is to add input and output graph lists to the statement, along with the regular graph lists. The idea is to use the regular graph lists to collect graphs for creation/deletion, and the input graph lists to collect graphs that act as a source of information (such as information queried in a WHERE clause) and the output graph lists to collect graphs that act as a destination of information (such as triples added in an INSERT or CONSTRUCT clause).

There are a couple of reasons for this change. Primarily, there needs to be a distinction between input and output graphs in a CONSTRUCT...WHERE BQL statement. Also, having the distinction helps to reason through the code.

### DIFF
--- a/bql/grammar/grammar_test.go
+++ b/bql/grammar/grammar_test.go
@@ -255,32 +255,12 @@ func TestRejectByParse(t *testing.T) {
 	}
 }
 
-func TestAcceptOpsByParseAndSemantic(t *testing.T) {
+func TestAcceptCreateAndDropGraphOpsByParseAndSemantic(t *testing.T) {
 	table := []struct {
 		query   string
 		graphs  int
 		triples int
 	}{
-		// Insert data.
-		{`insert data into ?a {/_<foo> "bar"@[1975-01-01T00:01:01.999999999Z] /_<foo>};`, 1, 1},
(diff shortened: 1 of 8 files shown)


bql/grammar/grammar_test.go
```diff
@@ -271,6 +301,31 @@ func TestAcceptOpsByParseAndSemantic(t *testing.T) {
...
+		if got, want := len(st.Data()), entry.triples; got != want {
+			t.Errorf("Parser.consume: Failed to collect right number of triples for case %v; got %d, want %d", entry, got, want)
+		}
+	}
+}
+
+func TestAcceptDeleteDataFromGraphOpsByParseAndSemantic(t *testing.T) {
+	table := []struct {
+		query   string
+		graphs  int
```
xllora: I'd rather suggest to use []string{"?graph",...} instead of testing for length. Here and on the other tests. Since this aso improves on correctness.

palimarrao: Done. Also added tests for the construct statement.

palimarrao:
PTAL, thanks!

PR #79, opened 2017-07-20
Implementation of the DECONSTRUCT statement.

palimarrao:


bql/grammar/grammar.go
```diff
@@ -101,6 +101,19 @@ func BQL() *Grammar {
 					NewTokenType(lexer.ItemSemicolon),
 				},
 			},
+			{
+				Elements: []Element{
+					NewTokenType(lexer.ItemDeconstruct),
+					NewSymbol("DECONSTRUCT_FACTS"),
+					NewTokenType(lexer.ItemAt),
```
xllora: into instead of at?

bql/grammar/grammar_test.go
```diff
@@ -319,14 +354,30 @@ func TestAcceptGraphOpsByParseAndSemantic(t *testing.T) {
 			 ?s "old_predicate_2"@[,] ?o2.
 			 ?s "old_predicate_3"@[,] ?o3};`, empty, []string{"?b"}, []string{"?a"}, 0},
```
xllora: Here and below, can you separate this into multiple lines?

palimarrao: Done. The current indentation is after gofmt. Wonder if it should be indented more to the left same col as the statement param?

bql/grammar/grammar.go
palimarrao: Changed to 'in' instead of 'at'.

bql/grammar/grammar_test.go
```diff
@@ -145,6 +145,18 @@ func TestAcceptByParse(t *testing.T) {
 		 from ?b where {?s "old_predicate_1"@[,] ?o1.
 				?s "old_predicate_2"@[,] ?o2.
 				?s "old_predicate_3"@[,] ?o3};`,
+		// Test Deconstruct clause.
+		`deconstruct {?s "new_predicate"@[] ?o} in ?a from ?b where {?s "old_predicate"@[,] ?o} having ?s = ?o;`,
+		`deconstruct {?s "new_predicate"@[] ?o} in ?a from ?b where {?s "old_predicate"@[,] ?o};`,
+		`deconstruct {?s ?p ?o.
+			      _:v "_subject"@[] ?s.
```
xllora: Mmh, I recall we discussed to start with the triple syntax with not reification. I thought that would include not supporting blank node syntax for deconstruct right now. Also not sure what the value for those would be since those are uniquely generated, likely it would not deconstruct anything.

palimarrao: Yep, you are right, DECONSTRUCT does not support reification. However, the reason I added support for blank node is that it is possible in the CONSTRUCT statement to explicitly have blank nodes as the subject for a triple. So while we cannot DECONSTRUCT reified blank nodes with this syntax (without knowing the random uuid), we can DECONSTRUCT explicit blank nodes with known uuids. However, I am fine removing this functionality since it could be confusing, but I think it has a valid use case.

xllora: I found this a bit misleading. The main reason in my head is the following. When you query in the where clause, you query blank nodes if you need to. This means, that after all, you get them like any other node/regular binding. Hence, for the deconstruct part you would actually use the regular binding you have obtained from the where  clause query. 

The blank node notation, needs to be consistent. If you have a blank node placeholder it must generate a new unique blank node. That would mostly mean, generating non preexisting nodes by definition, hence, never deleting anything. 

For the reasons above, I would suggest remove the _:v notation from the syntax of destruct.

palimarrao: That's a good point. I have removed blank node support.

xllora:
Thanks for the change @palimarrao

PR #91, opened 2019-10-18
Adding comparison of literals and bindings.

rbkloss:
The clause:
`HAVING ?capacity > "10"^^type:int64`
Was documented but not implemented, this PR should solve that.


bql/semantic/expression.go
```diff
@@ -87,6 +87,80 @@ type evaluationNode struct {
...
+		csEL, csER string
+	)
+
+	if lerr != nil && rerr != nil {
+		return false, fmt.Errorf("comparison operations require either the left or right operands to be a binding but neither had an existing row")
+	}
+
+	if lerr == nil {
+		csEL = cs(leftCell)
+	} else {
```
xllora: Given the code above, else will not be called. Also, see comment above about early return.

bql/semantic/expression.go
```diff
@@ -87,6 +87,80 @@ type evaluationNode struct {
...
+	}
+	if rerr == nil {
+		csER = cs(rightCell)
+	} else {
+		csER = e.rS
+	}
+
+	switch e.op {
+	case EQ:
+		return reflect.DeepEqual(csEL, csER), nil
```
xllora: You are just comparing strings, at this point, prefer == instead.

bql/semantic/expression_test.go
```diff
@@ -387,6 +387,27 @@ func TestNewEvaluator(t *testing.T) {
 			err:  false,
 			want: false,
 		},
+		{
+			id: "?foo = \"text\"^^type:text",
+			in: []ConsumedElement{
```
xllora: Could you also add a few test on operation correctness in https://github.com/google/badwolf/blob/master/bql/planner/planner_test.go

bql/semantic/expression.go
```diff
@@ -87,6 +87,80 @@ type evaluationNode struct {
...
+	}
+
+	cs := func(c *table.Cell) string {
+		if c.L != nil {
+			return strings.TrimSpace(c.L.ToComparableString())
+		}
+		return strings.TrimSpace(c.String())
+	}
+
+	leftCell, lerr := getLeft()
```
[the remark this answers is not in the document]
rbkloss: Hi X, in the if we are actually checking that both sides have no binding, when that happens we return since we need at least one binding to evaluate the comparison.

When there is one error, it mean that is the literal we want.

bql/semantic/expression.go
rbkloss: the error actually signals us that that is a literal and not a binding, so, in the if above we check if there are two literals, if that is the case we bail out.

below, in the else we assign csEL to the literal value when there is an error.

bql/semantic/expression.go
rbkloss: Done

bql/semantic/expression.go
```diff
@@ -87,6 +87,80 @@ type evaluationNode struct {
...
+
+	literalOnLeft bool
+
+	lS string
+	rS string
+}
+
+func (e *evaluationLiteralNode) Evaluate(r table.Row) (bool, error) {
+	// Binary evaluation
+	getLeft := func() (*table.Cell, error) {
```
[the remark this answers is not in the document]
rbkloss: Will Do

bql/semantic/expression.go
```diff
@@ -87,6 +87,80 @@ type evaluationNode struct {
 	rB string
 }
 
+type evaluationLiteralNode struct {
```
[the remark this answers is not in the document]
rbkloss: Done

bql/semantic/expression_test.go
rbkloss: Done

bql/semantic/expression.go
```diff
@@ -22,6 +22,7 @@ import (
 
 	"github.com/google/badwolf/bql/lexer"
 	"github.com/google/badwolf/bql/table"
+	litutils "github.com/google/badwolf/triple/literal"
```
xllora: Just curios. What is the goal of the alias? I do not recall we are using elsewhere in the codebase.

bql/semantic/expression.go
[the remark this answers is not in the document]
xllora: Mmh, OK.

xllora:
I cannot still merge it. Can you check the push request and update it?

PR #94, opened 2019-11-17
Switch to go modules

aerostitch:
Hi guys,

As it's been a bit over a year that the main projects have started to use the go modules, I figured that maybe it would be good to switch badwolf to use go modules too.
So here's the PR for it.

Note: it also switches travis to:
* use go modules
* test on go 1.12, 1.13 and tip (removes 1.11)
* adds the `-v -race` options to the tests to have better visibility and check for race conditions
* limit the depth of the tree cloning to 1 as we don't look at the history (makes the clone more efficient)

Thanks for your time,
Joseph

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,11 @@ language: go
 
 sudo: false
 
-env: GO111MODULE=on
+env:
+  - GO111MODULE=on
+
+git:
+  depth: 1
 
 go:
   - "1.12"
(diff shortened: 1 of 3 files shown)


aerostitch:
Cool, thanks @xllora .
I'm going to have some time to kill in the coming weeks so if you have other stuffs on the back burner I can definitely help, just let me know. I'll be more than happy to help. (I've got a few last cleanup to push and after I'll have a look at the open issues, see if I can help there)

I've seen that other Google projects have started to use Github actions instead of Travis. Let me know if that's something you'd be interested in.

xllora:
Thanks @aerostitch! 

Yes, actions would be something worth improving. Feel free to open an issue on it so we do not forget it even if you cannot work on it :) Same with the the open issues that have been open for a while, any help will be more than welcome :D

PR #96, opened 2019-11-22
Switch CI from travis to github actions

aerostitch:
Hi guys,

Sorry for the late PR I've been pretty sick those last few days.

This PR switches the run of the CI from Travis to Github actions.
It is a stripped version of what you can find in https://github.com/google/go-github/blob/master/.github/workflows/tests.yml

Note:
* specifying each folder in the go test command doesn't really change anything in our case so I switched back to `./...` but I can switch that back if you want.
* I added a profile coverprofile and covermode this waif you want to integrate with codecove one day as I've seen other Google projects do should be adding the secret in the project settings and 4 lines of yaml that can be copied from https://github.com/google/go-github/blob/master/.github/workflows/tests.yml#L45.
* As this creates the workflow from a fork by a user who is not a collaborator on the project, the workflow won't appear in the PR until you either merge it in master or in another branch within the repository itself I think. But they do appear in my fork: https://github.com/aerostitch/badwolf/actions

Thanks for your time,
Joseph


aerostitch:
This also fixes an issue found using go vet in a test file (missing keys on the fields of a composite literal).

aerostitch:
Thanks @xllora ! Yeah 5 days in the sun in Cabo helped the recovery! ;)
No worries about the speed of response, it's all good! ;)
Yeah actions are pretty powerful!
Tahnks for the review!
Joseph

PR #97, opened 2019-11-22
Remove double assignment of ctx in benchmarkQuery

aerostitch:
Hi guys,

I missed this last time I was looking at ineffassign.
ctx here is re-reclared on the next line so just cleaning up the first unused assignment.

Thanks,
Joseph


xllora:
Indeed! Thanks @aerostitch!

PR #98, opened 2019-11-22
Fix the string representation of a ConstructPredicateObjectPair and add more tests

aerostitch:
Hi guys,

In the current `ConstructPredicateObjectPair.String()` when we look for an object binding, whether `c.O` is nil or not we set object to true meaning that we just never enter the Predicate portion.

This will lead to only partially returning the representation of the object (leaving out the ID an the anchorBinding).

Plus in the predicate portion we already take care of printing the binding so basically it's doing twice the same thing.

Joseph

### DIFF
--- a/bql/semantic/semantic_test.go
+++ b/bql/semantic/semantic_test.go
@@ -455,6 +455,72 @@ func TestInputOutputBindings(t *testing.T) {
 	}
 }
 
+func TestConstructPredicateObjectPairString(t *testing.T) {
+	// Testing that NewNodeFromStrings is not the point of this package. Taking the example from the unit tests.
+	n, _ := node.NewNodeFromStrings("/some/type", "id_1")
+	// Testing NewImmutable is not the point of this package.
+	immutFoo, _ := predicate.NewImmutable("foo")
+	o := triple.NewNodeObject(n)
+
+	table := []struct {
+		pop  *ConstructPredicateObjectPair
+		want string
+	}{
+		{&ConstructPredicateObjectPair{}, `@[]][]`},
+		{
+			&ConstructPredicateObjectPair{
+				P:              immutFoo,
+				PID:            "?predID",
+				PBinding:       "?predBinding",
+				PAnchorBinding: "?predAnchorBinding",
+				PTemporal:      true,
+				O:              o,
+				OBinding:       "?objBinding",
+				OID:            "?objID",
+				OAnchorBinding: "?Popeyes",
+				OTemporal:      true,
+			},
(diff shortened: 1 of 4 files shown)


aerostitch:
During my tests I ended up finding a case where trying to do a `.String()` on a node created with just `Node{}` crashes the app so I added a fix and some tests for that but I can see to separate it in a separate PR if you want or just drop it too.

aerostitch:
Alright I mixed up `GraphClause` with `ConstructPredicateObjectPair` when making tests after dinner and started to write tests for `GraphClause.String()` but I guess that's not entirely lost. More test coverage! :)

aerostitch:
Alright, added the tests for `ConstructPredicateObjectPair.String()` too.

bql/semantic/semantic_test.go
```diff
@@ -96,6 +97,132 @@ func TestStatementAddData(t *testing.T) {
...
+	n, _ := node.NewNodeFromStrings("/some/type", "id_1")
+	// Testing NewImmutable is not the point of this package.
+	immutFoo, _ := predicate.NewImmutable("foo")
+	nO, _ := node.NewNodeFromStrings("/some/other/type", "id_2")
+	o := triple.NewNodeObject(nO)
+	table := []struct {
+		gc   *GraphClause
+		want string
+	}{
+		{&GraphClause{}, `{ opt=false @[][] }`},
```
xllora: For consistency, could you add new lines before and after?

bql/semantic/semantic_test.go
```diff
@@ -96,6 +97,132 @@ func TestStatementAddData(t *testing.T) {
...
+	// Testing that NewNodeFromStrings is not the point of this package. Taking the example from the unit tests.
+	n, _ := node.NewNodeFromStrings("/some/type", "id_1")
+	// Testing NewImmutable is not the point of this package.
+	immutFoo, _ := predicate.NewImmutable("foo")
+	nO, _ := node.NewNodeFromStrings("/some/other/type", "id_2")
+	o := triple.NewNodeObject(nO)
+	table := []struct {
+		gc   *GraphClause
+		want string
+	}{
```
xllora: For consistency with the rest of the files, would it be possible to explicitly name the fields instead of relying on position?

aerostitch: Good catch! Done.

bql/semantic/semantic_test.go
aerostitch: Done. Good catch, thanks!

aerostitch:
Thanks for your review!
I've made the suggested modifications. Let me know if there's more on this one. Happy to adapt if needed ;)

Joseph

PR #101, opened 2019-12-11
Fix syntax parsing of BEFORE, AFTER and BETWEEN clauses

aerostitch:
As @xllora and @rbkloss were discussing about in #90 , the current syntax for the AFTER block does not reflect what the doc says it's supposed to be.

In fact the problem extends to the BEFORE and BETWEEN clauses too.

This PR changes the lexer and hook to behave as expected for those 3 clauses.

Note: I haven't added unit tests specific to the new lexer function as I think the higher level tests at the grammar and planner level already catch that but I can add more tests there if needed. Just let me know.

Closes: #90

### DIFF
--- a/bql/lexer/lexer.go
+++ b/bql/lexer/lexer.go
@@ -340,15 +340,16 @@ type stateFn func(*lexer) stateFn
 
 // lexer holds the state of the scanner.
 type lexer struct {
-	input    string     // the string being scanned.
-	start    int        // start position of this item.
-	pos      int        // current position in the input.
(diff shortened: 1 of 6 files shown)


bql/lexer/lexer.go
```diff
@@ -376,6 +377,10 @@ func lexToken(l *lexer) stateFn {
 	for {
 		{
 			r := l.peek()
+			// special predicate parsing for global level timestamps
```
xllora: Nit: Prefer use full sentences for comments with proper punctuation.

PR #102, opened 2019-12-13
Adding comparison of literals and bindings.

rbkloss:
Creating this PR again since there was some unmergeable conflict in the last one [pull91](https://github.com/google/badwolf/pull/91).


xllora:
And fixed the issues :D

Thanks for the fix!

PR #116, opened 2020-07-15
Refactor - remove duplicate code for Expressions

rogerlucena:
Refactoring a little `expression.go` to remove some duplicate code, extracting functions that will be shared from different methods now and in the future, while also organizing the code to present itself in a more intuitive order.

### DIFF
--- a/bql/semantic/expression.go
+++ b/bql/semantic/expression.go
@@ -84,64 +84,12 @@ func (o OP) String() string {
 // evaluationNode represents the internal representation of one expression.
 type evaluationNode struct {
 	op OP
+
 	lB string
 	rB string
 }
 
-// comparisonForNodeLiteral represents the internal representation of a expression of comparison between a binding and a node literal.
-type comparisonForNodeLiteral struct {
-	op OP
-
-	lB  string
-	rNL string
-}
-
-func (e *comparisonForNodeLiteral) Evaluate(r table.Row) (bool, error) {
-	// Binary evaluation
-	eval := func() (*table.Cell, error) {
-		var (
-			eL *table.Cell
-			ok bool
-		)
-		eL, ok = r[e.lB]
-		if !ok {
-			return nil, fmt.Errorf("comparison operations require the binding value for %q for row %q to exist", e.lB, r)
-		}
-		return eL, nil
-	}
-
-	cs := func(c *table.Cell) string {
-		if c.L != nil {
-			return strings.TrimSpace(c.L.ToComparableString())
-		}
-		return strings.TrimSpace(c.String())
-	}
-
-	eL, err := eval()
-	if err != nil {
-		return false, err
-	}
-	csEL, csER := cs(eL), strings.TrimSpace(e.rNL)
-	switch e.op {
-	case EQ:
-		return reflect.DeepEqual(csEL, csER), nil
-	case LT:
-		return cs(eL) < csER, nil
-	case GT:
-		return csEL > csER, nil
-	default:
-		return false, fmt.Errorf("boolean evaluation require a boolean operation; found %q instead", e.op)
-	}
-}
-
-// comparisonForLiteral represents the internal representation of a expression of comparison between a literal and a binding.
-type comparisonForLiteral struct {
-	op OP
-
-	lS string
-	rS string
-}
-
+// format cell to a trimmed comparable string
 func formatCell(c *table.Cell) string {
 	if c.L != nil {
 		return strings.TrimSpace(c.L.ToComparableString())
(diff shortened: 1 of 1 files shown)


bql/semantic/expression.go
```diff
@@ -84,85 +84,41 @@ func (o OP) String() string {
...
-	switch e.op {
-	case EQ:
-		return reflect.DeepEqual(csEL, csER), nil
-	case LT:
-		return cs(eL) < csER, nil
-	case GT:
-		return csEL > csER, nil
-	default:
-		return false, fmt.Errorf("boolean evaluation require a boolean operation; found %q instead", e.op)
+// binary evaluation - get value for binding in given row
```
rbkloss, line 100: Extracting this so that it is no longer an anonymous function causes the name to be too broad.
I would suggest to rename it to explain the specific value it is extracting, e.g., cell cellFromTable.

bql/semantic/expression.go
```diff
@@ -84,85 +84,41 @@ func (o OP) String() string {
...
-		}
-		return eL, nil
-	}
-
-	cs := func(c *table.Cell) string {
-		if c.L != nil {
-			return strings.TrimSpace(c.L.ToComparableString())
-		}
-		return strings.TrimSpace(c.String())
+// format cell to a trimmed comparable string
```
rbkloss, line 92: I think you have to start the comment with the name of the function

bql/semantic/expression.go
```diff
@@ -84,85 +84,41 @@ func (o OP) String() string {
...
-			ok  bool
-		)
-		val, ok = r[binding]
-		if !ok {
-			return nil, fmt.Errorf("comparison operations require the binding value for %q for row %q to exist", binding, r)
-		}
-		return val, nil
-	}
-
 	csLit := func(lit string) (string, error) {
```
rbkloss, line 122: unrelated to your change, but since we are here, should we maybe rename this?
csLit does not mean much to the reader

bql/semantic/expression.go
```diff
@@ -197,6 +153,33 @@ func (e *comparisonForLiteral) Evaluate(r table.Row) (bool, error) {
 	}
 }
 
+// comparisonForNodeLiteral represents the internal representation of an expression of comparison between a binding and a node literal.
+type comparisonForNodeLiteral struct {
+	op OP
```
rbkloss, line 158: I wonder if we should add comments to the fields of these structs.
It is kind of hard to figure out what each field means without seeing them in being used in the code.

bql/semantic/expression.go
```diff
@@ -197,6 +153,33 @@ func (e *comparisonForLiteral) Evaluate(r table.Row) (bool, error) {
...
+		return false, err
+	}
+
+	csEL, csER := formatCell(eL), strings.TrimSpace(e.rNL)
+	switch e.op {
+	case EQ:
+		return reflect.DeepEqual(csEL, csER), nil
+	case LT:
+		return csEL < csER, nil
+	case GT:
```
rbkloss, line 176: I am beginning to wonder if we should have checked if the type of the binding is correct and return error if not.
WDYT roger?

bql/semantic/expression.go
```diff
@@ -441,8 +425,8 @@ func internalNewEvaluator(ce []ConsumedElement) (Evaluator, []ConsumedElement, e
 			return e, res, nil
 		}
 
-		if bndTkn.Type == lexer.ItemNode {
-			e, err := NewEvaluationExpressionForNodeLiteral(op, tkn.Text, bndTkn.Text)
+		if bndTkn.Type == lexer.ItemLiteral {
```
rbkloss, line 428: Out of curiosity, why change the ordering of the blocks?

bql/semantic/expression.go
rogerlucena: I agree, I have just renamed it, thank you Kloss.

bql/semantic/expression.go
rogerlucena: Done.

bql/semantic/expression.go
rogerlucena: I agree, I have just renamed it to `formatLiteral` as we named the other function `formatCell`.

bql/semantic/expression.go
rogerlucena: That is true, it is indeed, I have just added more comments for them.

bql/semantic/expression.go
rogerlucena: As I was navigation the code it seemed more natural the ordering `comparisonForLiteral` -> `comparisonForNodeLiteral` -> `comparisonForPredicateLiteral` (in a following PR). So I changed the ordering here inside `internalNewEvaluator` too to be coherent with that. I can also change back to the previous ordering if you prefer.

bql/semantic/expression.go
rogerlucena: I believe it makes sense too, I am adding these checks right now.

bql/semantic/expression.go
```diff
@@ -81,24 +81,60 @@ func (o OP) String() string {
...
+		return csEL > csER, nil
+	default:
+		return false, fmt.Errorf("boolean evaluation require a boolen operation; found %q instead", e.op)
 	}
-	return strings.TrimSpace(c.String())
 }
 
-// binary evaluation - get value for binding in given row
-func getValue(binding string, r table.Row) (*table.Cell, error) {
+// getCellFromRow retrives the value of a binding from a given row
```
rbkloss, line 136: Golang usually does not recommend prefixing  functions with get because it is argued that it adds nothing to the readability of the function.

PR #117, opened 2020-07-15
Fix support for comparison between ID/TYPE bindings and text literals inside HAVING clauses

rogerlucena:
Previously, comparisons between ID bindings and text literals inside HAVING clauses (such as below) were not effective:

```
SELECT ?s, ?predID, ?o
FROM ?family
WHERE {
  ?s ?p ID ?predID ?o
}
HAVING ?predID < "parent_of"^^type:text;
```

Now, they are fully supported (`<`, `>` and `=`), proceeding as a lexicographical comparison between the ID binding and the text literal that follows.

The same is now supported for bindings created using the TYPE keyword too.

Error handling for comparison expressions were also improved.

### DIFF
--- a/bql/planner/planner_test.go
+++ b/bql/planner/planner_test.go
@@ -767,30 +767,30 @@ func TestPlannerQuery(t *testing.T) {
 
 		st := &semantic.Statement{}
 		if err := p.Parse(grammar.NewLLk(entry.q, 1), st); err != nil {
-			t.Errorf("Parser.consume: failed to parse query %q with error: %v", entry.q, err)
+			t.Errorf("parser.Parse(%s)\n= %v; want nil error", entry.q, err)
 		}
 		plnr, err := New(ctx, s, st, 0, 10, nil)
 		if err != nil {
-			t.Errorf("planner.New failed to create a valid query plan with error: %v", err)
+			t.Errorf("planner.New() = _, %v; want nil error", err)
 		}
 		tbl, err := plnr.Execute(ctx)
 		if !entry.wantErr && err != nil {
-			t.Errorf("planner.Execute failed for query %q with error: %v", entry.q, err)
+			t.Errorf("planner.Execute(%s)\n= _, %v; want nil error", entry.q, err)
 			continue
 		}
 		if entry.wantErr && err == nil {
-			t.Errorf("planner.Execute for query %q should have returned an error, but did not", entry.q)
+			t.Errorf("planner.Execute(%s)\n= _, nil; want non-nil error", entry.q)
 			continue
 		}
 		if entry.wantErr {
-			// an error was expected and it was indeed caught above.
+			// an error was expected and it was indeed caught.
 			continue
 		}
 		if got, want := len(tbl.Bindings()), entry.nbs; got != want {
-			t.Errorf("tbl.Bindings returned the wrong number of bindings for %q; got %d, want %d", entry.q, got, want)
+			t.Errorf("planner.Execute(%s)\n= a Table with %d bindings; want %d", entry.q, got, want)
 		}
 		if got, want := len(tbl.Rows()), entry.nrws; got != want {
-			t.Errorf("planner.Execute failed to return the expected number of rows for query %q; got %d want %d\nGot:\n%v\n", entry.q, got, want, tbl)
+			t.Errorf("planner.Execute(%s)\n= a Table with %d rows; want %d\nTable:\n%v\n", entry.q, got, want, tbl)
 		}
 	}
 }
--- a/bql/planner/planner_test.go
+++ b/bql/planner/planner_test.go
@@ -760,19 +760,21 @@ func TestPlannerQuery(t *testing.T) {
 	s, ctx := memory.NewStore(), context.Background()
 	populateStoreWithTriples(ctx, s, "?test", testTriples, t)
 	for _, entry := range testTable {
+		// Test setup:
 		p, err := grammar.NewParser(grammar.SemanticBQL())
 		if err != nil {
 			t.Fatalf("grammar.NewParser should have produced a valid BQL parser but got error: %v", err)
 		}
-
 		st := &semantic.Statement{}
 		if err := p.Parse(grammar.NewLLk(entry.q, 1), st); err != nil {
-			t.Errorf("parser.Parse(%s)\n= %v; want nil error", entry.q, err)
+			t.Fatalf("parser.Parse failed for query \"%s\"\nwith error: %v", entry.q, err)
 		}
 		plnr, err := New(ctx, s, st, 0, 10, nil)
 		if err != nil {
-			t.Errorf("planner.New() = _, %v; want nil error", err)
+			t.Fatalf("planner.New failed to create a valid query plan with error: %v", err)
 		}
+
+		// Actual test:
 		tbl, err := plnr.Execute(ctx)
 		if !entry.wantErr && err != nil {
 			t.Errorf("planner.Execute(%s)\n= _, %v; want nil error", entry.q, err)
--- a/bql/planner/planner_test.go
+++ b/bql/planner/planner_test.go
@@ -566,6 +566,26 @@ func TestPlannerQuery(t *testing.T) {
 			nbs:  2,
 			nrws: 4,
 		},
+		{
+			q: `SELECT ?s_id, ?height
+					FROM ?test
+					WHERE {
+						?s ID ?s_id "height_cm"@[] ?height
+					}
+					HAVING ?s_id = "alice"^^type:text;`,
+			nbs:  2,
+			nrws: 1,
+		},
+		{
+			q: `SELECT ?p_id, ?o
+					FROM ?test
+					WHERE {
+						/u<peter> ?p ID ?p_id ?o
+					}
+					HAVING ?p_id < "parent_of"^^type:text;`,
+			nbs:  2,
+			nrws: 4,
+		},
 	}
 
 	s, ctx := memory.NewStore(), context.Background()
--- a/bql/planner/planner_test.go
+++ b/bql/planner/planner_test.go
@@ -707,6 +707,50 @@ func TestPlannerQuery(t *testing.T) {
 			nbs:  2,
 			nrws: 7,
 		},
+		{
+			q: `SELECT ?p, ?p_id, ?o
+				FROM ?test
+				WHERE {
+					/u<peter> ?p ID ?p_id ?o
+				}
+				HAVING ?p_id = "bought"^^type:text
+				AFTER 2016-03-01T00:00:00-08:00;`,
+			nbs:  3,
+			nrws: 2,
+		},
+		{
+			q: `SELECT ?p, ?p_id, ?o
+				FROM ?test
+				WHERE {
+					/u<peter> ?p ID ?p_id ?o
+				}
+				HAVING ?p_id < "parent_of"^^type:text
+				BEFORE 2016-03-01T00:00:00-08:00;`,
+			nbs:  3,
+			nrws: 3,
+		},
+		{
+			q: `SELECT ?p, ?p_id, ?o
+				FROM ?test
+				WHERE {
+					/u<peter> ?p ID ?p_id ?o
+				}
+				HAVING ?p_id = "bought"^^type:text
+				BETWEEN 2016-02-01T00:00:00-08:00, 2016-03-01T00:00:00-08:00;`,
+			nbs:  3,
+			nrws: 2,
+		},
+		{
+			q: `SELECT ?p, ?p_id, ?o
+				FROM ?test
+				WHERE {
+					/u<peter> ?p ID ?p_id ?o
+				}
+				HAVING ?p_id < "work_with"^^type:text
+				BEFORE 2016-02-01T00:00:00-08:00;`,
+			nbs:  3,
+			nrws: 4,
+		},
 	}
 
 	s, ctx := memory.NewStore(), context.Background()
--- a/bql/planner/planner_test.go
+++ b/bql/planner/planner_test.go
@@ -685,6 +685,28 @@ func TestPlannerQuery(t *testing.T) {
 			nbs:  2,
 			nrws: 4,
 		},
+		{
+			q: `SELECT ?s, ?s_type
+				FROM ?test
+				WHERE {
+					?s TYPE ?s_type ?p ?o
+				}
+				HAVING ?s_type = "/c"^^type:text;`,
+			err:  false,
+			nbs:  2,
+			nrws: 4,
+		},
+		{
+			q: `SELECT ?s, ?s_type
+				FROM ?test
+				WHERE {
+					?s TYPE ?s_type ?p ?o
+				}
+				HAVING ?s_type < "/l"^^type:text;`,
+			err:  false,
+			nbs:  2,
+			nrws: 7,
+		},
 	}
 
 	s, ctx := memory.NewStore(), context.Background()
--- a/bql/planner/planner_test.go
+++ b/bql/planner/planner_test.go
@@ -108,7 +108,7 @@ func insertAndDeleteTest(t *testing.T) {
                                /_<foo> "bar"@[] "yeah"^^type:text};`
 	p, err := grammar.NewParser(grammar.SemanticBQL())
 	if err != nil {
-		t.Errorf("grammar.NewParser: should have produced a valid BQL parser, %v", err)
+		t.Errorf("grammar.NewParser should have produced a valid BQL parser but got error: %v", err)
 	}
 	stm := &semantic.Statement{}
 	if err = p.Parse(grammar.NewLLk(bql, 1), stm); err != nil {
@@ -145,7 +145,7 @@ func insertAndDeleteTest(t *testing.T) {
 			      /_<foo> "bar"@[] "yeah"^^type:text};`
 	p, err = grammar.NewParser(grammar.SemanticBQL())
 	if err != nil {
-		t.Errorf("grammar.NewParser: should have produced a valid BQL parser, %v", err)
+		t.Errorf("grammar.NewParser should have produced a valid BQL parser but got error: %v", err)
 	}
 	stm = &semantic.Statement{}
 	if err = p.Parse(grammar.NewLLk(bql, 1), stm); err != nil {
@@ -194,7 +194,7 @@ func TestPlannerCreateGraph(t *testing.T) {
 	bql := `create graph ?foo, ?bar;`
 	p, err := grammar.NewParser(grammar.SemanticBQL())
 	if err != nil {
-		t.Errorf("grammar.NewParser: should have produced a valid BQL parser, %v", err)
+		t.Errorf("grammar.NewParser should have produced a valid BQL parser but got error: %v", err)
 	}
 	stm := &semantic.Statement{}
 	if err = p.Parse(grammar.NewLLk(bql, 1), stm); err != nil {
@@ -225,7 +225,7 @@ func TestPlannerDropGraph(t *testing.T) {
 	bql := `drop graph ?foo, ?bar;`
 	p, err := grammar.NewParser(grammar.SemanticBQL())
 	if err != nil {
-		t.Errorf("grammar.NewParser: should have produced a valid BQL parser")
+		t.Errorf("grammar.NewParser should have produced a valid BQL parser but got error: %v", err)
 	}
 	stm := &semantic.Statement{}
 	if err = p.Parse(grammar.NewLLk(bql, 1), stm); err != nil {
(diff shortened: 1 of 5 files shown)


bql/semantic/expression.go
```diff
@@ -90,11 +90,18 @@ type evaluationNode struct {
...
 
 // format cell to a trimmed comparable string
-func formatCell(c *table.Cell) string {
+func formatCell(c *table.Cell) (string, error) {
 	if c.L != nil {
-		return strings.TrimSpace(c.L.ToComparableString())
+		return strings.TrimSpace(c.L.ToComparableString()), nil
 	}
-	return strings.TrimSpace(c.String())
+	if c.S != nil {
```
rbkloss, line 97: Although I suggested to do this, I wonder if we could really always consider a string cell to be comparable to a text literal.
I think we will have to add extensive tests to check that there is no obscure usage that will break with this.

bql/semantic/expression.go
```diff
@@ -90,11 +90,18 @@ type evaluationNode struct {
...
+		formatted, err := literal.DefaultBuilder().Build(literal.Text, *c.S)
+		if err != nil {
+			return "", fmt.Errorf("could not build literal from string, received error: %v", err)
```
rbkloss, line 100: I wonder if we should annotate the error instead

bql/semantic/expression.go
```diff
@@ -136,11 +143,15 @@ func (e *comparisonForLiteral) Evaluate(r table.Row) (bool, error) {
...
 	)
 
-	csEL = formatCell(leftCell)
+	csEL, err = formatCell(leftCell)
+	if err != nil {
+		return false, err
+	}
 	csER, err = csLit(e.rS)
 	if err != nil {
 		return false, err
```
rbkloss, line 152: Here and in line 148 and line 139 I think we should annotate the error expanding on its message to explicitly tell that the error was obtained while evaluating a comparison of a binding and a literal

bql/semantic/expression.go
```diff
@@ -167,7 +178,12 @@ func (e *comparisonForNodeLiteral) Evaluate(r table.Row) (bool, error) {
...
 	}
 
-	csEL, csER := formatCell(eL), strings.TrimSpace(e.rNL)
+	csEL, err := formatCell(eL)
+	if err != nil {
+		return false, err
```
rbkloss, line 183: annotate the erro here, as well, please

bql/semantic/expression_test.go
```diff
@@ -15,11 +15,13 @@
...
+	"fmt"
 	"testing"
 
 	"github.com/google/badwolf/bql/lexer"
 	"github.com/google/badwolf/bql/table"
 	"github.com/google/badwolf/triple/literal"
+	"github.com/google/badwolf/triple/node"
 )
 
 func buildLiteral(l string) string {
```
rbkloss, line 27: Did you check if there is any usage for this, now?
If not, could you remove it, please?

bql/semantic/expression_test.go
```diff
@@ -412,11 +430,76 @@ func TestNewEvaluator(t *testing.T) {
...
+				}),
+			},
+			r: table.Row{
+				"?id": &table.Cell{S: table.CellString("abc")},
+			},
+			err:  false,
+			want: true,
+		},
+		{
+			id: "?s ID ?id < \"bbb\"^^type:text",
```
rbkloss, line 462: If the escaped quotes starts to bother you, it is possible to also write this as:
`?s ID ?id < "bbb"^^type:text`

bql/planner/planner_test.go
```diff
@@ -566,6 +566,26 @@ func TestPlannerQuery(t *testing.T) {
...
+		{
+			q: `SELECT ?p_id, ?o
+					FROM ?test
+					WHERE {
+						/u<peter> ?p ID ?p_id ?o
+					}
+					HAVING ?p_id < "parent_of"^^type:text;`,
+			nbs:  2,
+			nrws: 4,
+		},
```
rbkloss, line 588: Should we add tests comparing with a literal of a "incorrect" type?
Maybe also add a test with a node ID and a node literal?

bql/semantic/expression.go
rogerlucena: As discussed, `func Annotate(err error, msg string) error` not available out of Google3, proceeding with `fmt.Errorf` instead.

bql/semantic/expression.go
rogerlucena: Done.

bql/semantic/expression.go
rogerlucena: Totally agree. Done.

bql/semantic/expression_test.go
rogerlucena: The function `buildLiteral` is indeed not being used anymore, I have just removed it.

bql/semantic/expression_test.go
rogerlucena: Cool, I did not know about that. Thank you Kloss.

bql/planner/planner_test.go
rogerlucena: For now I added the tests for these two cases inside `expression_test.go`, since there we already had a structure to verify expected errors.
Should I add a similar structure inside `planner_test.go` too (like an `err` entry below `nrws` for `testTable`, inside `TestPlannerQuery`) and also repeat these tests there?

bql/semantic/expression.go
```diff
@@ -89,7 +89,7 @@ func formatCell(c *table.Cell) (string, error) {
 	if c.S != nil {
 		formatted, err := literal.DefaultBuilder().Build(literal.Text, *c.S)
 		if err != nil {
-			return "", fmt.Errorf("could not build literal from string, received error: %v", err)
+			return "", fmt.Errorf("in formatCell, could not build literal from string - received error: %v", err)
```
rbkloss, line 92: WDYT of:
could not build a literal from string while trying to formatCell, got error:

bql/semantic/expression.go
```diff
@@ -175,27 +175,26 @@ func (e *comparisonForLiteral) Evaluate(r table.Row) (bool, error) {
...
 		return false, fmt.Errorf("in comparisonForLiteral.Evaluate got error: %v", err)
 	}
-	csER, err = formatLiteral(e.rS)
+	if leftCell.S != nil && lit.Type() != literal.Text {
+		return false, fmt.Errorf("a string binding can only be compared with a literal of type text, got literal %v instead", lit.String())
+	}
+
+	var (
+		csEL, csER string // comparable string expressions left and right
+	)
```
rbkloss, line 192: WDYT of:
// comparable string expressions for left and right tokens.
var csEL, csER string

Do note that comments need to have punctuation but error message don't

bql/semantic/expression.go
```diff
@@ -222,6 +221,9 @@ func (e *comparisonForNodeLiteral) Evaluate(r table.Row) (bool, error) {
 	if err != nil {
 		return false, fmt.Errorf("in comparisonForNodeLiteral.Evaluate got error: %v", err)
 	}
+	if eL.N == nil {
+		return false, fmt.Errorf("only a node binding should be compared with a node literal, but got %v instead", eL.String())
```
rbkloss, line 225: This could also have a mix of correct and incorrect rows at runtime so I think we should log the error message and return false with no error.

Point of notice: maybe we should add these swallowed errors to the tracer and if something do go wrong the tracer could signal the swallowed bits or the client could request the trace as well, and check these kind of errors.

docs/bql.md
```diff
@@ -407,7 +407,7 @@ that would require multiple clauses and extra bindings:
 ```
 
 Note that the intervals defined by `before`, `after`, and `between` are always closed (the 
-limits of the intervals are included). You can then understand the `after` as a '>=', the 
+limits of the intervals are included). You can then understand the `after` as a `>=`, the
```
rbkloss, line 410: should this be in another commit/PR?

bql/semantic/expression.go
rogerlucena: Done.

bql/semantic/expression.go
```diff
@@ -175,27 +175,26 @@ func (e *comparisonForLiteral) Evaluate(r table.Row) (bool, error) {
...
 	}
 
-	csEL, err = formatCell(leftCell)
+	lit, err := literal.DefaultBuilder().Parse(e.rS)
 	if err != nil {
 		return false, fmt.Errorf("in comparisonForLiteral.Evaluate got error: %v", err)
 	}
-	csER, err = formatLiteral(e.rS)
+	if leftCell.S != nil && lit.Type() != literal.Text {
+		return false, fmt.Errorf("a string binding can only be compared with a literal of type text, got literal %v instead", lit.String())
```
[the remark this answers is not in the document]
rogerlucena: And what about bindings obtained with the TYPE operator? I am starting to think about them for some time now. It may be necessary to separate another PR for them just like I will be doing with the AT operator too.

bql/semantic/expression.go
rogerlucena: Done.

docs/bql.md
rogerlucena: I agree, thank you Kloss. I have just extracted this minor correction to a separate commit.

bql/semantic/expression_test.go
```diff
@@ -195,6 +189,22 @@ func TestBooleanEvaluationNode(t *testing.T) {
...
+	if err != nil {
+		panic("Could not parse text literal got err: " + err.Error())
+	}
+	return lit
+}
+
+func newNodeFromStringOrDie(nodeType, nodeID string) *node.Node {
+	n, err := node.NewNodeFromStrings(nodeType, nodeID)
+	if err != nil {
+		panic(fmt.Sprintf("Could not build node from type %s and value %s", nodeType, nodeID))
```
thiagovas, line 203:
```suggestion
		panic(fmt.Sprintf("Could not build node from type %q and value %q", nodeType, nodeID))
```

thiagovas:
@googlebot I fixed it.

bql/semantic/expression.go
rogerlucena: I see. I will let it to be added to the tracer then and remove this error from here, for now.

bql/semantic/expression.go
```diff
@@ -175,27 +175,26 @@ func (e *comparisonForLiteral) Evaluate(r table.Row) (bool, error) {
...
-	)
-
-	formatLiteral := func(lit string) (string, error) {
-		n, err := literal.DefaultBuilder().Parse(lit)
-		if err != nil {
-			return "", err
-		}
-		return n.ToComparableString(), nil
+	if leftCell.L == nil && leftCell.S == nil {
+		return false, fmt.Errorf("only a literal binding or a string binding (such as an ID binding) should be compared with a literal, but got %v instead", leftCell.String())
```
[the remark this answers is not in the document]
rogerlucena: I see. I will let this error to be added to the tracer then and remove it from here, for now.

bql/planner/planner_test.go
rogerlucena: I have just added a similar structure for checking expected errors inside `planner_test.go` too, and added additional tests there and inside `expression_test.go` for code coverage.

bql/semantic/expression.go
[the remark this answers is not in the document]
rogerlucena: I have just added some tests for comparisons with TYPE bindinds too.

bql/semantic/expression.go
rogerlucena: I have just added more tests, mainly having comparisons involving ID bindings in HAVING clauses followed by AFTER/BETWEEN/BEFORE clauses (as a way of scrutinizing this use case that may become commonly used in the future).

bql/planner/planner_test.go
rogerlucena: I have just added one more check for literal type coherence when comparing literals inside `expression.go`, along with some tests for that too.

bql/semantic/expression.go
```diff
@@ -81,37 +81,44 @@ func (o OP) String() string {
...
+func formatCell(c *table.Cell) (string, error) {
 	if c.L != nil {
-		return strings.TrimSpace(c.L.ToComparableString())
+		return strings.TrimSpace(c.L.ToComparableString()), nil
 	}
-	return strings.TrimSpace(c.String())
+	if c.S != nil {
+		formatted, err := literal.DefaultBuilder().Build(literal.Text, *c.S)
+		if err != nil {
+			return "", fmt.Errorf("could not build a literal from string while trying to formatCell, got error: %v", err)
```
Tati1701, line 92: You can add information on c.S so this error is more useful - i.e. help the maintainer to understand where the error happened too.

bql/semantic/expression.go
```diff
@@ -81,37 +81,44 @@ func (o OP) String() string {
...
+	// Binary evaluation.
 	eval := func() (*table.Cell, *table.Cell, error) {
 		var (
 			eL, eR *table.Cell
 			ok     bool
 		)
 		eL, ok = r[e.lB]
 		if !ok {
-			return nil, nil, fmt.Errorf("comparison operations require the binding value for %q for row %q to exist", e.lB, r)
+			return nil, nil, fmt.Errorf("comparison operation requires the binding value for %q for row %q to exist", e.lB, r)
```
Tati1701, line 117: `%q` should be used for strings, but `r` is a table.Row. Either use `%v` or convert it to string before (here and below).

bql/semantic/expression.go
```diff
@@ -120,7 +127,16 @@ func (e *evaluationNode) Evaluate(r table.Row) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	csEL, csER := formatCell(eL), formatCell(eR)
+
+	csEL, err := formatCell(eL)
+	if err != nil {
+		return false, fmt.Errorf("in evaluationNode.Evaluate got error: %v", err)
```
Tati1701, line 133: Shouldn't this message say that the error was formatting cell? And maybe which cell it was trying to format? 
Messages like`in evaluationNode.Evaluate` don't add much information.
Please apply the same rationale throughout this file.

bql/semantic/expression.go
```diff
@@ -129,54 +145,59 @@ func (e *evaluationNode) Evaluate(r table.Row) (bool, error) {
 	case GT:
 		return csEL > csER, nil
 	default:
-		return false, fmt.Errorf("boolean evaluation require a boolen operation; found %q instead", e.op)
+		return false, fmt.Errorf("boolean evaluation requires a boolean operation; found %q instead", e.op)
```
Tati1701, line 148: Same situation: `e.op` is an enum, and will not be "pretty printed". 
Example: https://play.golang.org/p/boTxJB5RRIw

bql/semantic/expression.go
```diff
@@ -129,54 +145,59 @@ func (e *evaluationNode) Evaluate(r table.Row) (bool, error) {
...
-	rS string // right string
+	lS string // left string.
+	rS string // right string.
 }
 
 func (e *comparisonForLiteral) Evaluate(r table.Row) (bool, error) {
 	leftCell, err := cellFromRow(e.lS, r)
 	if err != nil {
-		return false, err
+		return false, fmt.Errorf("in comparisonForLiteral.Evaluate got error: %v", err)
```
Tati1701, line 176: Similar comment: you got error extracting a cell from a row, right? What were you trying to extract? From where? Might these help the person debugging this to find the issue, along with the error message?

bql/semantic/expression_test.go
```diff
@@ -195,6 +189,22 @@ func TestBooleanEvaluationNode(t *testing.T) {
...
+
+func newNodeFromStringOrDie(nodeType, nodeID string) *node.Node {
+	n, err := node.NewNodeFromStrings(nodeType, nodeID)
+	if err != nil {
+		panic(fmt.Sprintf("Could not build node from type %q and value %q", nodeType, nodeID))
+	}
+	return n
+}
+
 func TestNewEvaluator(t *testing.T) {
```
Tati1701, line 208: The test struct you probably want something called `wantErr` rather than `err` to indicate whether you want an error or not (boolean field).

bql/semantic/expression_test.go
```diff
@@ -496,20 +592,150 @@ func TestNewEvaluator(t *testing.T) {
...
+				}),
+			},
+			r: table.Row{
+				"?foo": &table.Cell{L: buildLiteralOrDie(`"100.0"^^type:float64`)},
 			},
 			err:  false,
 			want: true,
 		},
 	}
 	for _, entry := range testTable {
```
Tati1701, line 727: You don't have a lot of setup to do in this test. I'd advise you split the test function into two to avoid the complexity you add here with all the ifs.

bql/semantic/expression_test.go
```diff
@@ -496,20 +592,150 @@ func TestNewEvaluator(t *testing.T) {
...
 		eval, err := NewEvaluator(entry.in)
-		if !entry.err && err != nil {
```
Tati1701, line 507: Please identify the function you are testing in the error messages, as well as the input and got/want.

For example, if you are testing the second return of `Evaluate` (the error) say:
`t.Errorf("e.Evaluate(%q) = _, %v; want nil", case.id, err)` // for a case where you didn't expect error but got one
And then when you proceed to comparing the first return:
`t.Errorf("e.Evaluate(%q) = %v, _; want %v, _", case.id, got, want)`
Note that these messages inform the maintainer what was the function under test, what was the input, the result you are looking at and what you expected. 

It would also be helpful, in this case, to print diffs so it is easier for the reader to compare got/want. For example:
`t.Errorf("e.Evaluate(%q) returned diff (-want +got):\n%s", case.id, diff)`
You can use a library such as `cmp.Diff` to do so:
https://godoc.org/github.com/google/go-cmp/cmp#Diff

bql/semantic/expression.go
rogerlucena: Very well pointed, I have just fixed it. Thank you Tati.

bql/semantic/expression.go
rogerlucena: Done. Thank you very much for the example.

bql/semantic/expression_test.go
rogerlucena: Done. Thank you very much for these insights Tati.

bql/semantic/expression.go
rogerlucena: Done. Thank you Tati.

bql/semantic/expression.go
rogerlucena: Done.

bql/semantic/expression.go
rogerlucena: Done.

bql/semantic/expression_test.go
rogerlucena: I will be using the next PR to separate the tests that expect an error from the tests that do not. I will put them in two different functions (for both `expression_test.go` and `planner_test.go`). Thank you again for the advice Tati.

bql/semantic/expression_test.go
rogerlucena: I will be doing this in the next PR. Thank you once again Tati.

bql/planner/planner_test.go
```diff
@@ -763,30 +767,30 @@ func TestPlannerQuery(t *testing.T) {
...
 		st := &semantic.Statement{}
 		if err := p.Parse(grammar.NewLLk(entry.q, 1), st); err != nil {
-			t.Errorf("Parser.consume: failed to parse query %q with error: %v", entry.q, err)
+			t.Errorf("parser.Parse(%s)\n= %v; want nil error", entry.q, err)
 		}
 		plnr, err := New(ctx, s, st, 0, 10, nil)
 		if err != nil {
-			t.Errorf("planner.New failed to create a valid query plan with error: %v", err)
+			t.Errorf("planner.New() = _, %v; want nil error", err)
 		}
```
Tati1701: I'm assuming what is under test here is the Execute() method. If so, are `parser.Parse` and `planner.New` needed as setup for the test? If they are part of the setup: 
- Could they be extracted to a test helper?
- Can the actual system under test continue if they fail (i.e. should this be Fatalf instead of Errorf)?

Maybe it's worth adding a comment here to the reader to help them differentiate what is actually being tested. You want to make it super easy to identify what is under test and what is part of supporting that.

Since you are not comparing neither `st` nor `plnr` with anything else here, I'm going to carry on assuming they are setup. 

I know I asked you to identify function, input, output in tests in general. However, for setups that may be harder to do because you just don't want them to fail so you can carry on with your *actual* test.

Reminder: your goal here is to help the maintainer to understand what's going on. So, in this case, a helpful (and valid) error message could be:
`t.Fatalf("Failed parsing query %q: %v", entry.q, err)`
`t.Fatalf("Failed creating query plan for %q: %v", entry.q, err)`

bql/semantic/expression_test.go
```diff
@@ -181,26 +181,26 @@ func TestBooleanEvaluationNode(t *testing.T) {
...
 	lit, err := literal.DefaultBuilder().Parse(textLiteral)
 	if err != nil {
-		panic(fmt.Sprintf("Could not parse text literal got err: %v", err.Error()))
+		panic(fmt.Sprintf("could not parse text literal %q, got error: %v", textLiteral, err))
 	}
 	return lit
 }
 
-func newNodeFromStringOrDie(nodeType, nodeID string) *node.Node {
+func mustBuildNodeFromStrings(nodeType, nodeID string) *node.Node {
```
Tati1701, line 200: Optional: Do you build a Node from anything else? If not, I'd remove the `FromStrings` suffix and simply call this function `mustBuildNode`.

bql/planner/planner_test.go
rogerlucena: Thank you very much once again Tati. I totally agree and have just proceeded so.

bql/semantic/expression_test.go
rogerlucena: It is also possible to create a node from the parameters `(t *Type, id *ID)`, so I will keep this function being called `mustBuildNodeFromStrings ` for clarity for now.

Issue #113, opened 2020-07-07
Program freezes when querying time anchor on datasets with mixed mutable/immutable triples


rogerlucena:
When using the keyword `AT` to extract the time anchor from predicates in datasets with mixed mutuable/immutable triples, the program freezes and does not respond the query probably due to a deadlock or infinite loop.

To illustrate, given the graph `family` with the data below:

```
INSERT DATA INTO ?family {
  /u<joe> "parent_of"@[] /u<peter> .
  /u<peter> "height_cm"@[2006-01-02T15:04:05.999999999Z] "172"^^type:int64 .
  /u<joe> "height_cm"@[2006-01-02T15:04:05.999999999Z] "174"^^type:int64
};
```

When executing the following query the program freezes:

```
SELECT ?time
FROM ?family
WHERE {
  /u<joe> ?p AT ?time ?o
};
```

But, if you execute the same query on the graph without inserting the immutable triple `/u<joe> "parent_of"@[] /u<peter>`, it returns normally the expected result:

```
Result:
?time
2006-01-02T15:04:05.999999999Z
```

This Issue may be related with another one: the Issue https://github.com/google/badwolf/issues/40. It may be possibly related to the driver implementation as well.

You can use the code in the file below to reproduce the above results.

[at_keyword_freezing.pdf](https://github.com/google/badwolf/files/4882426/at_keyword_freezing.pdf)

rogerlucena:
This Issue was fixed with PR #119.

PR #119, opened 2020-07-24
Fix bug of TYPE bindings with non-node objects and AT bindings with immutable triples (Issue #113)

rogerlucena:
Previously, when extracting the TYPE of an object through a TYPE binding, you would find a **deadlock** and the program would freeze if you had objects that were not nodes in your graph. 

For example, in the graph `?family` with the following triples:

```
INSERT DATA INTO ?family {
  /u<peter> "height_cm"@[2006-01-02T15:04:05.999999999Z] "172"^^type:int64 .
  /u<joe> "height_cm"@[2006-01-02T15:04:05.999999999Z] "174"^^type:int64 .
  /u<joe> "parent_of"@[] /u<peter>
};
```

If you executed the following query:

```
SELECT ?o_type
FROM ?family
WHERE {
 /u<joe> ?p ?o TYPE ?o_type
};
```

You would have a deadlock due to the fact that your query would try to extract the TYPE from a literal object (eg: `"174"^^type:int64` here), that does not have a TYPE as in the case of a node object. It would also happen if you had objects that were predicates in your graph (like in the case of a reification). 

This PR comes to fix this problem.

It also fixes the problem of a deadlock when your query tries to extract the time anchor through AT bindings and you have immutable triples in your graph. This later problem with AT bindings and immutable triples is detailed in the issue https://github.com/google/badwolf/issues/113.

### DIFF
--- a/bql/planner/data_access.go
+++ b/bql/planner/data_access.go
@@ -570,7 +570,8 @@ func tripleToRow(t *triple.Triple, cls *semantic.GraphClause) (table.Row, error)
 	}
 	if cls.PAnchorBinding != "" {
 		if p.Type() != predicate.Temporal {
-			return nil, fmt.Errorf("cannot retrieve the time anchor value for non temporal predicate %q in binding %q", p, cls.PAnchorBinding)
+			// in the case of time anchor binding (eg: "bought"@[?time]) for an immutable predicate we just want (for now) to skip this triple and proceed to the next one, not returning any errors.
+			return nil, nil
 		}
 		t, err := p.TimeAnchor()
 		if err != nil {
--- a/bql/planner/data_access.go
+++ b/bql/planner/data_access.go
@@ -409,32 +409,26 @@ func simpleFetch(ctx context.Context, gs []storage.Graph, cls *semantic.GraphCla
 // table. The semantic graph clause is also passed to be able to identify what
 // bindings to set.
 func addTriples(ts <-chan *triple.Triple, cls *semantic.GraphClause, tbl *table.Table) error {
-	var lastErr error
-	for t := range ts {
-		if lastErr != nil {
-			// Drain the channel to avoid leaking goroutines.
-			continue
-		}
+	shouldIgnoreTriple := func(t *triple.Triple) (bool, error) {
 		if cls.PID != "" {
 			// The triples need to be filtered.
 			if string(t.Predicate().ID()) != cls.PID {
-				continue
+				return true, nil
 			}
 			if cls.PTemporal {
 				if t.Predicate().Type() != predicate.Temporal {
-					continue
+					return true, nil
 				}
 				ta, err := t.Predicate().TimeAnchor()
 				if err != nil {
-					lastErr = fmt.Errorf("failed to retrieve time anchor from time predicate in triple %s with error %v", t, err)
-					continue
+					return true, fmt.Errorf("failed to retrieve time anchor from time predicate in triple %s with error %v", t, err)
 				}
 				// Need to check the bounds of the triple.
 				if cls.PLowerBound != nil && cls.PLowerBound.After(*ta) {
-					continue
+					return true, nil
 				}
 				if cls.PUpperBound != nil && cls.PUpperBound.Before(*ta) {
-					continue
+					return true, nil
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


bql/planner/data_access.go
```diff
@@ -458,12 +465,17 @@ func addTriples(ts <-chan *triple.Triple, cls *semantic.GraphClause, tbl *table.
...
 		}
 		if r != nil {
 			tbl.AddRow(r)
 		}
 	}
+	if lastErr != nil {
+		return lastErr
+	}
+
 	return nil
```
thiagovas, line 479:
```suggestion
	return lastErr
```

rogerlucena: Done.

bql/planner/data_access.go
```diff
@@ -409,7 +409,12 @@ func simpleFetch(ctx context.Context, gs []storage.Graph, cls *semantic.GraphCla
 // table. The semantic graph clause is also passed to be able to identify what
 // bindings to set.
 func addTriples(ts <-chan *triple.Triple, cls *semantic.GraphClause, tbl *table.Table) error {
+	var lastErr error
 	for t := range ts {
+		if lastErr != nil {
+			// Drain the channel to avoid leaking goroutines.
+			continue
+		}
```
rogerlucena, line 417: **rbkloss** commented:

I think we can simplify this by always iterating on the ts channel after the loop which would effectively drain the channel.
It would be then your option of ignoring the error altogether or checking if it is non nil and breaking the loop to the draining one. Maybe even extract the channel draining logic to a DrainChannel function.

Specifically I think a break, instead of a continue will make readability easier.

bql/planner/data_access.go
```diff
@@ -570,10 +580,10 @@ func tripleToRow(t *triple.Triple, cls *semantic.GraphClause) (table.Row, error)
 			return nil, nil
 		}
 	}
-
 	if cls.PAnchorAlias != "" {
 		if p.Type() != predicate.Temporal {
-			return nil, fmt.Errorf("cannot retrieve the time anchor value for non temporal predicate %q in binding %q", p, cls.PAnchorAlias)
+			// in the case of AT binding for an immutable predicate we just want (for now) to skip this triple and proceed to the next one, not returning any errors.
```
jlsotomayorm, line 585: nit: the "for now" might be confusing because doesn't specify a fixed time period. Does it means that this behavior is temporal? Should we add a TODO with more explanation for a next PR or maybe just drop the "for now"?

bql/planner/data_access.go
jlsotomayorm: +1 . Not sure if this same as Ricardo's suggestion but indeed refactoring the 'draining'/skipping logic would indeed make the code more readable. In pseudocode we could have something like
for..
 ignore, err := IgnoreTriple()
 if ignore {
  continue
}
if err != nil {
  lastError = err
  continue
}
AddRow

bql/planner/data_access.go
rogerlucena: Hi Jose, thank you for the comments. Yes, it means that this behavior is temporal and it is indeed changed together with this comment in a next PR, that I have just finished (PR #127). 
But I agree that it is simpler and clearer to just drop this `"(for now)"` from this comment here at the moment. I have just done so. 🙂

bql/planner/data_access.go
rogerlucena: Thank you for the comment @jlsotomayorm. 
I have just mixed both of your comments to refactor the function `addTriples`, creating a new local boolean function `skipTriple` there and also separating the draining of the channel in a second loop below the original one as @rbkloss suggested.

bql/planner/data_access.go
```diff
@@ -409,62 +409,80 @@ func simpleFetch(ctx context.Context, gs []storage.Graph, cls *semantic.GraphCla
 // table. The semantic graph clause is also passed to be able to identify what
 // bindings to set.
 func addTriples(ts <-chan *triple.Triple, cls *semantic.GraphClause, tbl *table.Table) error {
-	for t := range ts {
+	shouldIgnoreTriple := func(t *triple.Triple) (bool, error) {
```
jlsotomayorm, line 412: nit: Although this function is only used in the context of addTriples, it is hindering readability (breaking the flow of "adding Triple logic)/ Let's move it to the package level. This also helps to make this function pure (in the sense that it won't be able to depend on variables from the outter scope) wdyt?

bql/planner/data_access.go
```diff
@@ -409,62 +409,80 @@ func simpleFetch(ctx context.Context, gs []storage.Graph, cls *semantic.GraphCla
...
+		if ignoreTriple {
+			continue
+		}
+
 		r, err := tripleToRow(t, cls)
 		if err != nil {
-			return err
+			break
 		}
 		if r != nil {
```
jlsotomayorm, line 476: super nit: in Go, it's more common to have the invalid/error logic nested. To make the code more standard you can always revert this logic.

if r == nil {
 // return error or error handling
}

// Normal logic -- adding row

bql/planner/data_access.go
rogerlucena: When writing `shouldIgnoreTriple` I thought about it but also thought that the argument of keeping this function local (as it is used only by `addTriples`) was stronger.
Thank you once again for the input Jose, I am new to Go so I appreaciate a lot all these hints. I do agree that with `shouldIgnoreTriple` in the package level we have a much more readable `addTriples`. 👍 
Then, I have just done as you suggested. 😀

bql/planner/data_access.go
rogerlucena: Thank you once again for the observation Jose.
I have just done so. 👍

PR #123, opened 2020-07-30
Refactor tests into separate functions that expect errors or not

rogerlucena:
For clarity, it is useful to not mix in the same test function cases for which it is expected an error and cases for which it is not. 

Separating these cases into two different test functions makes the code cleaner and avoids the unnecessary use of more complicated `ifs` inside the body of the test function. Also, it makes it easier to add new test cases in the future as it is simpler to realize where they belong to in the test file.

Then, this PR comes to follow this guideline for two important test functions inside `expression_test.go` and `planner_test.go`.

### DIFF
--- a/bql/planner/planner_test.go
+++ b/bql/planner/planner_test.go
@@ -272,286 +272,239 @@ func populateStoreWithTriples(ctx context.Context, s storage.Store, gn string, t
 
 func TestPlannerQuery(t *testing.T) {
 	testTable := []struct {
-		q       string
-		wantErr bool
-		nbs     int
-		nrws    int
+		q    string
+		nbs  int
+		nrws int
 	}{
 		{
-			q:       `select ?s, ?p, ?o from ?test where {?s ?p ?o};`,
-			wantErr: false,
-			nbs:     3,
-			nrws:    len(strings.Split(testTriples, "\n")) - 1,
+			q:    `select ?s, ?p, ?o from ?test where {?s ?p ?o};`,
+			nbs:  3,
+			nrws: len(strings.Split(testTriples, "\n")) - 1,
 		},
 		{
-			q:       `select ?s as ?s1, ?p as ?p1, ?o as ?o1 from ?test where {?s ?p ?o};`,
-			wantErr: false,
-			nbs:     3,
-			nrws:    len(strings.Split(testTriples, "\n")) - 1,
+			q:    `select ?s as ?s1, ?p as ?p1, ?o as ?o1 from ?test where {?s ?p ?o};`,
+			nbs:  3,
+			nrws: len(strings.Split(testTriples, "\n")) - 1,
 		},
 		{
-			q:       `select ?p, ?o from ?test where {/u<joe> ?p ?o};`,
-			wantErr: false,
-			nbs:     2,
-			nrws:    2,
+			q:    `select ?p, ?o from ?test where {/u<joe> ?p ?o};`,
+			nbs:  2,
+			nrws: 2,
 		},
 		{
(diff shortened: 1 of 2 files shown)


bql/semantic/expression_test.go
```diff
@@ -636,84 +619,102 @@ func TestNewEvaluator(t *testing.T) {
...
+			t.Fatalf("test %q should have never failed when creating a new evaluator from %v, got error: %v", entry.id, entry.in, err)
+		}
+
+		got, err := eval.Evaluate(entry.r)
+		if err != nil {
+			t.Errorf("%s: eval.Evaluate(%v) = _, %v; want nil error", entry.id, entry.r, err)
+		}
+		if want := entry.want; got != want {
+			t.Errorf("%s: eval.Evaluate(%v) = %v, _; want %v, _", entry.id, entry.r, got, want)
+		}
```
Tati1701, line 657: This function is called `TestNewEvaluator`, which indicates that it is testing the `NewEvaluator` func. However, it seems it is actually testing `eval.Evaluate` and thus should be named something like `TestEvaluatorEvaluate` or something in those lines. 
Is that the case or there's something I'm not seeing here?

bql/planner/planner_test.go
```diff
@@ -776,18 +691,10 @@ func TestPlannerQuery(t *testing.T) {
...
 			t.Errorf("planner.Execute(%s)\n= _, %v; want nil error", entry.q, err)
 			continue
 		}
-		if entry.wantErr && err == nil {
-			t.Errorf("planner.Execute(%s)\n= _, nil; want non-nil error", entry.q)
-			continue
-		}
-		if entry.wantErr {
-			// an error was expected and it was indeed caught.
-			continue
```
Tati1701, line 789: So good to see this gone! Thanks for the refactor! :D

bql/planner/planner_test.go
```diff
@@ -797,6 +704,53 @@ func TestPlannerQuery(t *testing.T) {
 	}
 }
 
+func TestPlannerQueryError(t *testing.T) {
+	testTable := []struct {
+		q string
+	}{
```
Tati1701, line 710: Question to you: is there something in the error you want to test? For example that you return the expected message, or that it has some other property (e.g. message contains something you're looking for, error is annotated with a particular status...)?
It's fine if you don't, but if you do this would be the place to put it :)

bql/planner/planner_test.go
rogerlucena: hahaha :D

bql/semantic/expression_test.go
rogerlucena: I agree with you and I have just renamed the function and its correspondent error checking one. Thank you once again Tati. 😃

bql/planner/planner_test.go
rogerlucena: Not for now, as I discussed here with Thiago, but we will keep an eye on this for the future. Thank you again for the suggestion Tati.

bql/planner/planner_test.go
```diff
@@ -272,286 +272,239 @@ func populateStoreWithTriples(ctx context.Context, s storage.Store, gn string, t
 
 func TestPlannerQuery(t *testing.T) {
 	testTable := []struct {
-		q       string
-		wantErr bool
-		nbs     int
-		nrws    int
+		q    string
+		nbs  int
```
rbkloss, line 276: change this name to numBindings/nBindings?

rogerlucena: Done.

PR #125, opened 2020-08-06
Refactor "data_access_test.go" for clarity and conciseness

rogerlucena:
Refactor `data_access_test.go` to make it cleaner, avoid repeated code (and multiple `if` clauses) and also use more clear variable names as a way of easing the job of anyone that may pass through or work on this file in the future.

### DIFF
--- a/bql/planner/data_access_test.go
+++ b/bql/planner/data_access_test.go
@@ -349,8 +349,7 @@ func TestDataAccessBasicBindings(t *testing.T) {
 		// Actual test:
 		r, err := tripleToRow(tpl, cls)
 		if err != nil {
-			t.Errorf(`tripleToRow(for triple "%s" and clause %q) got error: %v; want nil error`, tpl, cls, err)
-			continue
+			t.Fatalf(`tripleToRow(for triple "%s" and clause %q) got error: %v; want nil error`, tpl, cls, err)
 		}
 		bindings := []string{"?s", "?p", "?o"}
 		entryCells := []*table.Cell{entry.sBinding, entry.pBinding, entry.oBinding}
@@ -397,8 +396,7 @@ func TestDataAccessTripleToRowSubjectBindings(t *testing.T) {
 		// Actual test:
 		r, err := tripleToRow(tpl, entry.cls)
 		if err != nil {
-			t.Errorf(`tripleToRow(for triple "%s" and clause %q) got error: %v; want nil error`, tpl, entry.cls, err)
-			continue
+			t.Fatalf(`tripleToRow(for triple "%s" and clause %q) got error: %v; want nil error`, tpl, entry.cls, err)
 		}
 		bindings := []string{"?s", "?alias", "?type", "?id"}
 		entryCells := []*table.Cell{entry.sBinding, entry.sAlias, entry.sTypeAlias, entry.sIDAlias}
@@ -453,8 +451,7 @@ func TestDataAccessTripleToRowPredicateBindings(t *testing.T) {
 		// Actual test:
 		r, err := tripleToRow(tpl, entry.cls)
 		if err != nil {
-			t.Errorf(`tripleToRow(for triple "%s" and clause %q) got error: %v; want nil error`, tpl, entry.cls, err)
-			continue
+			t.Fatalf(`tripleToRow(for triple "%s" and clause %q) got error: %v; want nil error`, tpl, entry.cls, err)
 		}
 		bindings := []string{"?p", "?alias", "?id", "?anchorBinding", "?anchorAlias"}
 		entryCells := []*table.Cell{entry.pBinding, entry.pAlias, entry.pIDAlias, entry.pAnchorBinding, entry.pAnchorAlias}
@@ -523,8 +520,7 @@ func TestDataAccessTripleToRowObjectBindings(t *testing.T) {
 		// Actual test:
 		r, err := tripleToRow(tpl, entry.cls)
 		if err != nil {
-			t.Errorf(`tripleToRow(for triple "%s" and clause %q) got error: %v; want nil error`, tpl, entry.cls, err)
-			continue
+			t.Fatalf(`tripleToRow(for triple "%s" and clause %q) got error: %v; want nil error`, tpl, entry.cls, err)
 		}
 		bindings := []string{"?o", "?alias", "?type", "?id", "?anchorBinding", "?anchorAlias"}
 		entryCells := []*table.Cell{entry.oBinding, entry.oAlias, entry.oTypeAlias, entry.oIDAlias, entry.oAnchorBinding, entry.oAnchorAlias}
@@ -574,8 +570,7 @@ func TestDataAccessTripleToRowObjectBindingsDropping(t *testing.T) {
 		// Actual test:
 		r, err := tripleToRow(tpl, entry.cls)
 		if err != nil {
-			t.Errorf(`tripleToRow(for triple "%s" and clause %q) got error: %v; want nil error`, tpl, entry.cls, err)
-			continue
+			t.Fatalf(`tripleToRow(for triple "%s" and clause %q) got error: %v; want nil error`, tpl, entry.cls, err)
 		}
 		if r != nil {
 			t.Errorf(`tripleToRow(for triple "%s" and clause %q) failed to drop triple and returned row: %v; want nil row`, tpl, entry.cls, r)
--- a/bql/planner/data_access_test.go
+++ b/bql/planner/data_access_test.go
@@ -17,6 +17,7 @@ package planner
 import (
 	"context"
 	"errors"
+	"fmt"
 	"reflect"
 	"sync"
 	"testing"
@@ -319,19 +320,19 @@ func TestDataAccessBasicBindings(t *testing.T) {
 		oBinding *table.Cell
 	}{
 		{
-			t:        n.String() + "\t" + p.String() + "\t" + n.String(),
+			t:        fmt.Sprintf("%s\t%s\t%s", n, p, n),
 			sBinding: &table.Cell{N: n},
 			pBinding: &table.Cell{P: p},
 			oBinding: &table.Cell{N: n},
 		},
 		{
-			t:        n.String() + "\t" + p.String() + "\t" + p.String(),
+			t:        fmt.Sprintf("%s\t%s\t%s", n, p, p),
 			sBinding: &table.Cell{N: n},
 			pBinding: &table.Cell{P: p},
 			oBinding: &table.Cell{P: p},
 		},
 		{
-			t:        n.String() + "\t" + p.String() + "\t" + l.String(),
+			t:        fmt.Sprintf("%s\t%s\t%s", n, p, l),
 			sBinding: &table.Cell{N: n},
 			pBinding: &table.Cell{P: p},
 			oBinding: &table.Cell{L: l},
@@ -372,7 +373,7 @@ func TestDataAccessTripleToRowSubjectBindings(t *testing.T) {
 		sIDAlias   *table.Cell
 	}{
 		{
-			t: n.String() + "\t" + p.String() + "\t" + n.String(),
+			t: fmt.Sprintf("%s\t%s\t%s", n, p, n),
 			cls: &semantic.GraphClause{
 				SBinding:   "?s",
 				SAlias:     "?alias",
@@ -426,7 +427,7 @@ func TestDataAccessTripleToRowPredicateBindings(t *testing.T) {
 		pAnchorAlias   *table.Cell
 	}{
 		{
-			t: n.String() + "\t" + p.String() + "\t" + n.String(),
+			t: fmt.Sprintf("%s\t%s\t%s", n, p, n),
 			cls: &semantic.GraphClause{
 				PBinding:       "?p",
 				PAlias:         "?alias",
@@ -483,7 +484,7 @@ func TestDataAccessTripleToRowObjectBindings(t *testing.T) {
 		oAnchorAlias   *table.Cell
 	}{
 		{
-			t: n.String() + "\t" + p.String() + "\t" + n.String(),
+			t: fmt.Sprintf("%s\t%s\t%s", n, p, n),
 			cls: &semantic.GraphClause{
 				OBinding:   "?o",
 				OAlias:     "?alias",
@@ -496,7 +497,7 @@ func TestDataAccessTripleToRowObjectBindings(t *testing.T) {
 			oIDAlias:   &table.Cell{S: table.CellString(n.ID().String())},
 		},
 		{
-			t: n.String() + "\t" + p.String() + "\t" + p.String(),
+			t: fmt.Sprintf("%s\t%s\t%s", n, p, p),
 			cls: &semantic.GraphClause{
 				OBinding:       "?o",
 				OAlias:         "?alias",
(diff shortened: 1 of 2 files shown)


bql/planner/data_access.go
```diff
@@ -409,7 +409,12 @@ func simpleFetch(ctx context.Context, gs []storage.Graph, cls *semantic.GraphCla
 // table. The semantic graph clause is also passed to be able to identify what
 // bindings to set.
 func addTriples(ts <-chan *triple.Triple, cls *semantic.GraphClause, tbl *table.Table) error {
+	var lastErr error
 	for t := range ts {
+		if lastErr != nil {
+			// Drain the channel to avoid leaking goroutines.
+			continue
+		}
```
rbkloss, line 417: I think we can simplify this by always iterating on the ts channel after the loop which would effectively drain the channel.
It would be then your option of ignoring the error altogether or checking if it is non nil and breaking the loop to the draining one. Maybe even extract the channel draining logic to a DrainChannel function.

Specifically I think a break, instead of a continue will make readability easier.

bql/planner/data_access_test.go
```diff
@@ -407,58 +407,56 @@ func TestDataAccessTripleToRowSubjectBindings(t *testing.T) {
 
 func TestDataAccessTripleToRowPredicateBindings(t *testing.T) {
 	n, p, _ := testNodeTemporalPredicateLiteral(t)
-	ts, err := p.TimeAnchor()
+	ta, err := p.TimeAnchor()
```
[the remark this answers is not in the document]
thiagovas: +1

bql/planner/data_access_test.go
```diff
@@ -407,58 +407,56 @@ func TestDataAccessTripleToRowSubjectBindings(t *testing.T) {
...
 	testTable := []struct {
-		t   string
-		cls *semantic.GraphClause
-		bc  *table.Cell
-		ac  *table.Cell
-		ic  *table.Cell
-		tc  *table.Cell
-		atc *table.Cell
+		t              string
+		cls            *semantic.GraphClause
```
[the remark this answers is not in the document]
thiagovas: +1

bql/planner/planner_test.go
```diff
@@ -670,6 +670,16 @@ func TestPlannerQuery(t *testing.T) {
...
+		{
+			q: `SELECT ?o, ?o_type
+				FROM ?test
+				WHERE {
+					?s ?p ?o TYPE ?o_type
+				}
+				HAVING (?s = /u<joe>) OR (?s = /l<barcelona>) OR (?s = /u<alice>);`,
+			nBindings: 2,
+			nRows:     2,
+		},
```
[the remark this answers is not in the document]
rogerlucena: This change is in fact added with the PR #119 ("Fix bug of TYPE bindings with non-node objects and AT bindings with immutable triples"), appearing here because of the rebase.
The intent of this test there is to ensure that we are not having a deadlock when trying to extract the `TYPE` from objects that are literals or predicates, as it was the case before. At this moment the expected behavior is to skip these triples, which will be changed with PR #126 later.
For that, the simplest way I found to test the `TYPE` extraction against all the three types of objects (nodes, predicates and literals) without querying the entire graph and returning 23 rows was restricting the `?s` to the three above. This way it returns only 2 rows and it is easier for anyone looking at the test to look at the graph and verify that the query returns the expected number of rows.

bql/planner/planner_test.go
```diff
@@ -680,6 +680,15 @@ func TestPlannerQuery(t *testing.T) {
...
 		},
+		{
+			q: `SELECT ?p, ?time, ?o
+				FROM ?test
+				WHERE {
+					/u<peter> ?p AT ?time ?o
+				};`,
+			nBindings: 3,
+			nRows:     4,
+		},
```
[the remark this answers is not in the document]
rogerlucena: In the case of error the test prints the Table with the bindings and then, at this moment, it is useful to see `?p` and `?o` as well (in addition to `?time`) to understand the context and which triples were being queried when the error was found.

bql/planner/data_access.go
rogerlucena: This change is in fact added with the PR #119 ("Fix bug of TYPE bindings with non-node objects and AT bindings with immutable triples"), appearing here because of the rebase.
I transferred this comment to the conversation there on the PR #119 there then.

bql/planner/data_access_test.go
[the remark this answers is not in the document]
rogerlucena: This `ta` meaning time anchor is used in a number of files in BW. I discussed with Thiago a bit more and we believe that if we decide to rename it we could let this for a follow up refactor PR.

bql/planner/data_access_test.go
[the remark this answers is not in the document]
rogerlucena: This `cls` meaning `clause` is used in a number of files in BW. I discussed with Thiago a bit more and we believe that if we decide to rename it we could let this for a follow up refactor PR.

bql/planner/data_access.go
```diff
@@ -409,7 +409,12 @@ func simpleFetch(ctx context.Context, gs []storage.Graph, cls *semantic.GraphCla
...
 		if cls.PID != "" {
```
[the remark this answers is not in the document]
rogerlucena: In the definition of `GraphClause` in `semantic.go` we have fields for subject (like `SBinding`, `SAlias`, `SID` ...), predicate (like `PBinding`, `PAlias`, `PID` ...) and object (like `OBinding`, `OAlias`, `OID` ...) that seem clear in this context and are used in a number of files in BW. I discussed with Thiago a bit more and we believe that if we decide to rename them we should indeed let it for a follow up refactor PR.

bql/planner/data_access_test.go
```diff
@@ -311,61 +311,65 @@ func TestDataAccessBasicBindings(t *testing.T) {
...
+		sBinding *table.Cell
+		pBinding *table.Cell
+		oBinding *table.Cell
 	}{
 		{
-			t:  n.String() + "\t" + p.String() + "\t" + n.String(),
-			sc: &table.Cell{N: n},
-			pc: &table.Cell{P: p},
-			oc: &table.Cell{N: n},
+			t:        n.String() + "\t" + p.String() + "\t" + n.String(),
```
Tati1701, line 322: You can do this by using the `fmt` package in a more concise way:
`fmt.Sprintf("%s\t%s\t%s", n, p, n)`
`%s` should call the `String` method for you :D

Please do that here and below :)

bql/planner/data_access_test.go
```diff
@@ -311,61 +311,65 @@ func TestDataAccessBasicBindings(t *testing.T) {
...
+			oBinding: &table.Cell{L: l},
 		},
 	}
+
 	for _, entry := range testTable {
+		// Setup for test:
 		tpl, err := triple.Parse(entry.t, literal.DefaultBuilder())
 		if err != nil {
-			t.Errorf("triple.Parse failed to parse valid triple %q with error %v", entry.t, err)
+			t.Fatalf(`triple.Parse failed to parse valid triple "%s" with error: %v`, entry.t, err)
```
Tati1701, line 345: Avoid quoting the string yourself (`"%s"`). `%q` does that for you :)

Tati1701, line 345: nit: this message is a bit verbose, and most of it doesn't add much to the reader. Does `Failed to parse %q: %v` work for you? It seems to convey the same in a more concise way.

bql/planner/data_access_test.go
```diff
@@ -311,61 +311,65 @@ func TestDataAccessBasicBindings(t *testing.T) {
...
 		r, err := tripleToRow(tpl, cls)
 		if err != nil {
-			t.Errorf("tripleToRow for triple %q and clasuse %v, failed with error %v", tpl, cls, err)
-		}
-		if got, want := r["?s"], entry.sc; !reflect.DeepEqual(got, want) {
-			t.Errorf("tripleToRow failed to return right value for binding \"?s\"; got %q, want %q", got, want)
-		}
-		if got, want := r["?p"], entry.pc; !reflect.DeepEqual(got, want) {
-			t.Errorf("tripleToRow failed to return right value for binding \"?p\"; got %q, want %q", got, want)
+			t.Errorf(`tripleToRow for triple "%s" and clause %q failed with error: %v`, tpl, cls, err)
```
Tati1701, line 351: Please identify the function, input, got and want.
Avoid quoting the string yourself, prefer to use `%q`.
This error message is very verbose. You can convey the same in a more concise, to-the-point way (suggestion below).
And avoid coupling `t.Error` followed by `continue`. In this case, if you can't proceed, please use `t.Failf` instead.
```
t.Fatalf(`tripleToRow(%q, %v) = _, %v; want _, nil`, entry.t, cls, err)
```
I'm not sure the type of `cls` so I suggested `%v`. 
Since you have the first value as string in the test, I suggested to use it instead of the parsed value, since it is probably more readable to the person debugging it.
If both inputs here are too long to be readable when debugging, then prefer adding a description to your test struct, and printing that instead :)

bql/planner/data_access_test.go
```diff
@@ -311,61 +311,65 @@ func TestDataAccessBasicBindings(t *testing.T) {
...
+			continue
 		}
-		if got, want := r["?o"], entry.oc; !reflect.DeepEqual(got, want) {
-			t.Errorf("tripleToRow failed to return right value for binding \"?o\"; got %q, want %q", got, want)
+		bindings := []string{"?s", "?p", "?o"}
+		entryCells := []*table.Cell{entry.sBinding, entry.pBinding, entry.oBinding}
+		for i, binding := range bindings {
+			if got, want := r[binding], entryCells[i]; !reflect.DeepEqual(got, want) {
+				t.Errorf(`tripleToRow(%q) = "%s"; want "%s"`, binding, got, want)
```
Tati1701, line 358: Identify that you're comparing the first output only (note that I added the `_` to indicate it below`). And again the `%q` suggestion :) 
```
t.Errorf(`tripleToRow(%q) = %q, _ ; want %q, _`, binding, got, want)
```

bql/planner/data_access_test.go
```diff
@@ -562,38 +560,43 @@ func TestDataAccessTripleToRowObjectBindingsDroping(t *testing.T) {
...
-			atc: &table.Cell{T: ts},
+			oBinding:       &table.Cell{P: p},
+			oAlias:         &table.Cell{P: p},
+			oIDAlias:       &table.Cell{S: table.CellString(string(p.ID()))},
+			oAnchorBinding: &table.Cell{T: ta},
+			oAnchorAlias:   &table.Cell{T: ta},
 		},
 	}
+
 	for _, entry := range testTable {
```
Tati1701, line 585: I'm not sure what I 'm missing here, but you seem to use only `entry.t`, and `entry.cls`. Where do you use the cells in the entry?

bql/planner/data_access_test.go
rogerlucena: In this case I did it this way because `entry.t` is a string with tabs `\t` inside. Then, when using `%q` and breaking the error to see how it appears for the user I saw some `"\t"` written there, which is not good for reading.
With `%s` it does not appear as `"\t"`  but as tab spaces (as we want), that is why I proceeded this way.

bql/planner/data_access_test.go
rogerlucena: Thank you for the comment Tati, I have just done so. 😀

bql/planner/data_access_test.go
rogerlucena: You are right. I believe that when this part of the code was first developed just a copy-paste was done from the setup of the test function above.
I have just removed what is not used there in `TestDataAccessTripleToRowObjectBindingsDropping` to make it cleaner. 👍

bql/planner/data_access_test.go
rogerlucena: It seems a bit verbose for me too. I have just made it more concise, thank you Tati. 🙂

bql/planner/data_access_test.go
rogerlucena: This was already changed in a later commit to comply with the standard `tripleToRow(%q) = _, %v; want ...` for tests. I believe you saw this older version here because you were reviewing commit by commit, right?
About the `%q`, I explained in a comment above but I proceeded this way because, after breaking the error to see how it would like for the user, I saw that `%q` prints tabs `\t` as brute `"\t"` strings, which is not good to make the error message more readable when debugging.

bql/planner/data_access_test.go
rogerlucena: Very well pointed, thank you for this observation Tati, I had indeed let it pass and now I have just proceeded as you pointed.
About the `%q`, I explained in a comment above but I proceeded this way because, after breaking the error to see how it would look like for the user, I saw that `%q` prints tabs `\t` as brute `"\t"` strings, which is not good to make the error message more readable when debugging.

bql/planner/data_access_test.go
rogerlucena: Then, if in a given error message I use `"%s"` for one variable because of this problem but `%q` for a second one because it does not have tabs `\t` inside, should I let it this way or use `"%s"` for both of them to keep a pattern?

Tati1701: I see. There's probably a way for rendering this in a better way, but I'm not familiar. You can keep it as is, although it is not a great solution. I'll see if I can find something and let you know :)

bql/planner/data_access_test.go
Tati1701: Sorry I let the other one unnoticed, thanks for updating to `t.Fatalf`.

The way it is now, this is not a case description, but stating the inputs in a verbose way. What I mean by case description is: What path does this exercise? Why did you choose this triple/clause combination?

If the inputs are small enough that the inputs are readable, prefer to do so in a more concise manner like I suggested earlier:

```
t.Fatalf(`tripleToRow("%s", %s) = _, %v; want _, nil`, entry.t, cls, err)
```

bql/planner/data_access_test.go
```diff
@@ -446,20 +445,19 @@ func TestDataAccessTripleToRowPredicateBindings(t *testing.T) {
 		// Setup for test:
 		tpl, err := triple.Parse(entry.t, literal.DefaultBuilder())
 		if err != nil {
-			t.Fatalf(`triple.Parse failed to parse valid triple "%s" with error: %v`, entry.t, err)
+			t.Fatalf(`triple.Parse failed for triple "%s", got error: %v`, entry.t, err)
```
Tati1701, line 448: The reader already knows this is a failure/error (because the test failed). You can be a bit more concise in your message:

```
t.Fatalf(`triple.Parse failed for triple "%s": %v`, entry.t, err)
```

rogerlucena: Done. Thank you Tati. 🙂

bql/planner/data_access_test.go
rogerlucena: Thank you Tati. 🙂

bql/planner/data_access_test.go
rogerlucena: As we discussed in the Google chat, I decided by the end to just use the standard concise pattern for error messages in tests. Thank you again for the patience and all the replies in the chat Tati. 🙂

rogerlucena:
Thank you, Tati! =]

PR #127, opened 2020-08-06
Adjust behavior of AT and time anchor bindings for immutable predicates

rogerlucena:
We have two types of time bindings: `AT` bindings (using the `AT` keyword) and time anchor bindings (eg: `"bought"@[?time]`). 

In both cases, if a graph pattern tries to extract the time anchor from an immutable predicate (like `"parent_of"@[]`) we want to skip this triple and do not show it in the query result (as it does not precisely match the exact graph pattern specified - the pattern required a time anchor). 

Nevertheless, there is one exception: if this graph pattern is marked as `OPTIONAL` we expect to see `<NULL>` in the query result for that time binding.

This PR comes to enforce this behavior.

To illustrate for the case of a `AT` binding, take the query below:

```
SELECT ?p, ?time, ?o
FROM ?test
WHERE {
    /u<peter> ?p AT ?time ?o
};
```

In a `?test` graph having only the two following triples:

```
/u<peter> "bought"@[2016-01-01T00:00:00-08:00] /c<mini>
```

We expect as result:

```
Result:
?p					?time				?o
"bought"@[2016-01-01T00:00:00-08:00]	2016-01-01T00:00:00-08:00	/c<mini>
```

Note how the triple with the immutable predicate `"parent_of"@[]` was skipped and not shown in the query result as it did not precisely match the exact graph pattern specified in the `WHERE` clause (it had no time anchor to be extracted by the `AT` binding).

```
SELECT ?p, ?time, ?o
FROM ?test
WHERE {
    /u<peter> ?p ?o .
    OPTIONAL { /u<peter> ?p AT ?time ?o }
};
```

We would expect as result:

```
Result:
?p					?time				?o
"parent_of"@[]				<NULL>				/u<eve>
"bought"@[2016-01-01T00:00:00-08:00]	2016-01-01T00:00:00-08:00	/c<mini>
```

Note how the triple with the immutable predicate `"parent_of"@[]` appeared in the query result this time, with the `?time` binding indeed shown as `<NULL>` in the result table. This happened because the clause with the `AT` binding was marked as `OPTIONAL` in the `WHERE` clause.

For time anchor bindings such as `"parent_of"@[?time]` the expected behavior enforced by this PR is analogous, as better detailed by the tests added in `planner_test.go`.

Note that this shall apply if we are extracting time bindings from objects that are immutable predicates too. For example, in a reification scenario like in the clause below (taken from a `WHERE`):

```
/_<blank_node> "_predicate"@[] ?obj AT ?time_obj
```

In the case of `?obj` being an immutable predicate such as `"parent_of"@[]` we also expect the triple to be skipped (as it did not precisely match the exact graph pattern detailed above). On the other hand, if this clause was marked as `OPTIONAL` we would expect `?time_obj` to appear as `<NULL>` in the query result.

For any other doubts on the expected behavior for a given situation refer to the new tests added in `planner_test.go` and `data_access_test.go`.

### DIFF
--- a/bql/planner/data_access.go
+++ b/bql/planner/data_access.go
@@ -423,7 +423,7 @@ func shouldIgnoreTriple(t *triple.Triple, cls *semantic.GraphClause) (bool, erro
 		if string(t.Predicate().ID()) != cls.PID {
 			return true, nil
 		}
-		if cls.PTemporal {
+		if cls.PTemporal && cls.PAnchorBinding == "" {
 			if t.Predicate().Type() != predicate.Temporal {
 				return true, nil
 			}
@@ -596,15 +596,17 @@ func tripleToRow(t *triple.Triple, cls *semantic.GraphClause) (table.Row, error)
 		}
 	}
 	if cls.PAnchorBinding != "" {
+		var c *table.Cell
 		if p.Type() != predicate.Temporal {
-			// in the case of time anchor binding (eg: "bought"@[?time]) for an immutable predicate we just want to skip this triple and proceed to the next one, not returning any errors.
-			return nil, nil
+			// in the case of time anchor bindings (eg: "bought"@[?time]) for immutable predicates we provide an empty Cell as we want <NULL> to appear in the query result.
+			c = &table.Cell{}
+		} else {
+			t, err := p.TimeAnchor()
+			if err != nil {
+				return nil, fmt.Errorf("failed to retrieve the time anchor value for predicate %q in binding %q with error: %v", p, cls.PAnchorBinding, err)
+			}
+			c = &table.Cell{T: t}
 		}
-		t, err := p.TimeAnchor()
-		if err != nil {
-			return nil, fmt.Errorf("failed to retrieve the time anchor value for predicate %q in binding %q with error %v", p, cls.PAnchorBinding, err)
-		}
-		c := &table.Cell{T: t}
 		r[cls.PAnchorBinding] = c
 		if !validBinding(cls.PAnchorBinding, c) {
 			return nil, nil
--- a/bql/planner/data_access.go
+++ b/bql/planner/data_access.go
@@ -716,12 +716,15 @@ func tripleToRow(t *triple.Triple, cls *semantic.GraphClause) (table.Row, error)
 				return nil, &skippableError{"cls.OAnchorAlias in non-optional clause", err}
 			}
 			c = &table.Cell{}
+		} else if p.Type() != predicate.Temporal {
+			// in the case of AT bindings for objects that are immutable predicates we provide an empty Cell as we want <NULL> to appear in the query result.
+			c = &table.Cell{}
 		} else {
-			ts, err := p.TimeAnchor()
+			t, err := p.TimeAnchor()
 			if err != nil {
-				return nil, err
+				return nil, fmt.Errorf("failed to retrieve the time anchor value for predicate %q in binding %q with error: %v", p, cls.OAnchorAlias, err)
 			}
-			c = &table.Cell{T: ts}
+			c = &table.Cell{T: t}
 		}
 		r[cls.OAnchorAlias] = c
 		if !validBinding(cls.OAnchorAlias, c) {
--- a/bql/planner/data_access.go
+++ b/bql/planner/data_access.go
@@ -611,15 +611,17 @@ func tripleToRow(t *triple.Triple, cls *semantic.GraphClause) (table.Row, error)
 		}
 	}
 	if cls.PAnchorAlias != "" {
+		var c *table.Cell
 		if p.Type() != predicate.Temporal {
-			// in the case of AT binding for an immutable predicate we just want to skip this triple and proceed to the next one, not returning any errors.
-			return nil, nil
+			// in the case of AT bindings for immutable predicates we provide an empty Cell as we want <NULL> to appear in the query result.
+			c = &table.Cell{}
+		} else {
+			t, err := p.TimeAnchor()
+			if err != nil {
+				return nil, fmt.Errorf("failed to retrieve the time anchor value for predicate %q in binding %q with error: %v", p, cls.PAnchorAlias, err)
+			}
+			c = &table.Cell{T: t}
 		}
-		t, err := p.TimeAnchor()
-		if err != nil {
-			return nil, fmt.Errorf("failed to retrieve the time anchor value for predicate %q in binding %q with error %v", p, cls.PAnchorAlias, err)
-		}
-		c := &table.Cell{T: t}
 		r[cls.PAnchorAlias] = c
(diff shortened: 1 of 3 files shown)


bql/planner/data_access.go
```diff
@@ -409,7 +409,12 @@ func simpleFetch(ctx context.Context, gs []storage.Graph, cls *semantic.GraphCla
 // table. The semantic graph clause is also passed to be able to identify what
 // bindings to set.
 func addTriples(ts <-chan *triple.Triple, cls *semantic.GraphClause, tbl *table.Table) error {
+	var lastErr error
 	for t := range ts {
+		if lastErr != nil {
+			// Drain the channel to avoid leaking goroutines.
+			continue
+		}
```
rbkloss: maybe do that change where we move this to after the loop and use a brak to get there, instead of continue.

rogerlucena: It appeared here again just because this branch `time-binding-behavior` is rebased over the not yet merged branch `fix-type-object` of a previous PR (that is passing Go readability now). But the refactor we did using break and moving the draing logic to a moment after the loop is also here too (see commit "Refactor the function "addTriples" to make it clearer" above).
These commits of the previous PR appear again here because of the way GitHub manages the PRs without looking at the previous ones or letting us identify other PRs that must be merged before the current one. It has been annoying me a little too, but I do not know another way at the moment to work around it.

thiagovas (approved):

bql/planner/data_access.go
```diff
@@ -523,7 +523,7 @@ func objectToCell(o *triple.Object) (*table.Cell, error) {
 	return nil, fmt.Errorf("unknown object type in object %q", o)
 }
 
-// tripleToRow converts a triple into a row using the binndings specidfied
+// tripleToRow converts a triple into a row using the bindings specified
 // on the graph clause.
```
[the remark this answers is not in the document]
rogerlucena: Thank you Thiago, I have just fixed that too. =]

bql/planner/data_access.go
```diff
@@ -595,30 +595,40 @@ func tripleToRow(t *triple.Triple, cls *semantic.GraphClause) (table.Row, error)
...
+				return nil, &skippableError{"cls.PAnchorBinding in non-optional clause", fmt.Errorf("tried to extract a time anchor binding from an immutable predicate")}
+			}
+			c = &table.Cell{}
+		} else {
+			t, err := p.TimeAnchor()
+			if err != nil {
+				return nil, fmt.Errorf("failed to retrieve the time anchor value for predicate %q in binding %q with error: %v", p, cls.PAnchorBinding, err)
+			}
+			c = &table.Cell{T: t}
 		}
```
jlsotomayorm, line 611: suggestion: This part of the code is already very indented. We can avoid one indentation if we avoid the else and always initialize the cell. Like:

c: = &table.Cell{}
if type != Temporal && !cls.Optional{
 return 'non-optional clause error'
}
if type == Temporal{
// set the 'T' on the cell
}
// We handle the table cell

This way we make clear our intend of when is the error path and when we fill the 'cell' struct.

bql/planner/data_access.go
```diff
@@ -595,30 +595,40 @@ func tripleToRow(t *triple.Triple, cls *semantic.GraphClause) (table.Row, error)
...
+			// In the case of AT bindings for immutable predicates we skip the triple if the clause is not optional, otherwise we provide an empty Cell as we want <NULL> to appear in the query result.
+			if !cls.Optional {
+				return nil, &skippableError{"cls.PAnchorAlias in non-optional clause", fmt.Errorf("tried to extract an AT binding from an immutable predicate")}
+			}
+			c = &table.Cell{}
+		} else {
+			t, err := p.TimeAnchor()
+			if err != nil {
+				return nil, fmt.Errorf("failed to retrieve the time anchor value for predicate %q in binding %q with error: %v", p, cls.PAnchorAlias, err)
+			}
```
jlsotomayorm, line 629: same as above

bql/planner/data_access_test.go
```diff
@@ -463,12 +485,49 @@ func TestDataAccessTripleToRowPredicateBindings(t *testing.T) {
...
+			t.Fatalf(`triple.Parse failed for triple "%s": %v`, entry.t, err)
+		}
+
+		// Actual test:
+		_, err = tripleToRow(tpl, entry.cls)
+		if _, ok := err.(*skippableError); !ok {
+			t.Errorf(`tripleToRow("%s", %q) = _, %v; want _, skippableError`, tpl, entry.cls, err)
+		}
+	}
+}
```
jlsotomayorm, line 522: For subtests, go has a Run method that allow us to distinguish between the subtests when the test fails.

Here's an example:

func TestGroupedParallel(t *testing.T) {
    for _, tc := range testCases {
        tc := tc // capture range variable
        t.Run(tc.Name, func(t *testing.T) {
            if got := foo(tc.in); got != tc.out {
                t.Errorf("got %v; want %v", got, tc.out)
            }
            ...
        })
    }
}

More info here: https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks

This allows you to define a name for each test case and t.Run will print it separately for you when one of the subtests fails.

rogerlucena: Thank you for the hint Jose, It is cool that Go has this, I did not know about that.
For the example above with the separation between "setup" with `t.Fatalf` and "actual test" with `t.Errof` along with the given error messages I believe it is clear what each test is about while also providing the tools for the user to understand where exactly was the error (in the case it happened), but it would be nice indeed to have a name for each test case as well to complement that (using `t.Run`).
This current version above is following the standard for test functions in this file and the other files in BadWolf, so I will let it like this for now but I will keep the `t.Run` in mind for a future overall refactor (to include names for each test case, for this and the other test functions as well).

jlsotomayorm (approved):

bql/planner/data_access.go
rogerlucena: We discussed and decided, for now, to let it as it is as it makes more immediate for who is reading to understand the logic of the behavior in each case detailed in the PR description (for `cls.OAnchorBinding` for example the logic is: 1. if the object is non-predicate; 2. if it is an immutable predicate; 3. if it is a temporal predicate (as expected); adjusting the behavior when appropriate taking into account if the clause is optional or not).

But these hints commented here will also be taken into account for future situations to simplify and make complex `if else` statements more compact.

bql/planner/data_access.go
rogerlucena: Same as previous comment.

bql/planner/data_access_test.go
rogerlucena: Issue created to track this: https://github.com/google/badwolf/issues/152.

Issue #124, opened 2020-07-31
Expand support of TYPE extraction for literals and predicates too, in addition to nodes


rogerlucena:
Currently, the TYPE keyword can be used to extract types only from nodes. In the case of a node like `/u<paul>` it would extract `/u`, for example.

But, as a feature request, it would be interesting to be able to extract "types" from predicates and literals as well, and also be able to use them in HAVING clauses to filter our results.

To illustrate, it would be nice that BQL supported queries such as:

```
SELECT ?s, ?p, ?o, ?o_type
FROM ?test
WHERE {
  ?s ?p ?o TYPE ?o_type
}
HAVING ?o_type = ^^type:float64;
```

On which it would understand that we want triples that have an object that is a literal, and that this literal is specifically a `float64` literal.

In a similar way, we could use `^^type:int64` for `int64` literals and `^^type:text` for `text` literals, and so on.

For predicates, the HAVING clause could be:

```
HAVING ?o_type = ^^type:predicate;
```
(useful in the case of reification and blank nodes)

Or even:

```
HAVING ?o_type = ^^type:predicate:immutable;
```

To specifically query for immutable predicates. We could analogically use `^^type:predicate:temporal` to specifically query for temporal predicates as well.

In the case of nodes, the HAVING clause could be:

```
HAVING ?o_type = ^^type:/u;
```

If we wanted only node objects with type `/u` (the users in our database).

This feature would be useful in a number of scenarios, as illustrated. 

Also, for coherence, it is important to remember that, in SELECT queries, the result for TYPE bindings should also follow the same pattern as above. To illustrate, the query:

```
SELECT ?o, ?o_type
FROM ?test
WHERE {
  ?s ?p ?o TYPE ?o_type
};
```

Should now return something like:

```
Result:
?o			?o_type
/u<paul>		^^type:/u
"3.7"^^type:float64	^^type:float64
"hello"^^type:text    	^^type:text
"parent_of"[]    	^^type:predicate:immutable 	
```

For consistency.

xllora:
Thanks for putting this explanation together @rogerlucena. I was wondering if you have settled on using `=` vs the `is`,`covariant`, `contravariant` combo alternatives for the comparison primitives.

rogerlucena:
Following Xavier's comment, an alternative to using `=` is implementing a `is` keyword to be used when comparing types. For example:

```
HAVING ?o_type is ^^type:predicate:immutable;
```

[...]

PR #128, opened 2020-08-06
Adjust behavior of TYPE bindings for non-node objects

rogerlucena:
In the case of `TYPE` bindings for non-node objects we want to skip this triple and do not show it in the query result as it does not precisely match the exact graph pattern specified (the pattern required a `TYPE` extraction, therefore the object had to be a node).

Nevertheless, there is one exception: if this graph pattern is marked as `OPTIONAL` we expect to see `<NULL>` in the query result for that `TYPE` binding.

This PR, then, comes to enforce this behavior.

In other words, if the object is a predicate or a literal we will not succeed in extracting the node type using the `TYPE` keyword and so we skip the triple if the clause is not optional, otherwise we put a `<NULL>` in the result table (for now, see Issue https://github.com/google/badwolf/issues/124). 

To illustrate, given a query like:

```
SELECT ?o, ?o_type
FROM ?test
WHERE {
    ?s ?p ?o TYPE ?o_type
};
```

In a `?test` graph with only the following triples:

```
/u<alice> "height_cm"@[] "174"^^type:int64
```

We expect as result:

```
Result:
?o			?o_type
/u<eve>			/u
```

Note how the triple with the non-node object `"174"^^type:int64` was skipped and not shown in the query result as it did not precisely match the exact graph pattern specified in the `WHERE` clause (the `TYPE` extraction made it necessary for the object to be a node).

```
SELECT ?o, ?o_type
FROM ?test
WHERE {
    ?s ?p ?o .
    OPTIONAL { ?s ?p ?o TYPE ?o_type }
};
```

We would expect as result:

```
Result:
?o			?o_type
"174"^^type:int64	<NULL>
/u<eve>			/u
```

Note how the triple with the literal object `"174"^^type:int64` appeared in the query result this time, with the `?o_type` binding indeed shown as `<NULL>` in the result table. This happened because the clause with the `TYPE` binding was marked as `OPTIONAL` in the `WHERE` clause.

For the case on which the objects are predicates (like in a reification scenario) the expected behavior enforced by this PR is analogous, as better detailed by the tests added by this PR in `planner_test.go` and `data_access_test.go`.

### DIFF
--- a/bql/planner/data_access.go
+++ b/bql/planner/data_access.go
@@ -649,12 +649,17 @@ func tripleToRow(t *triple.Triple, cls *semantic.GraphClause) (table.Row, error)
 		}
 	}
 	if cls.OTypeAlias != "" {
+		var c *table.Cell
 		n, err := o.Node()
 		if err != nil {
-			// in the case of TYPE binding for a non-node object we just want to skip this triple and proceed to the next one, not returning any errors.
-			return nil, nil
+			// In the case of TYPE bindings for non-node objects we skip the triple if the clause is not optional, otherwise we provide an empty Cell as we want <NULL> to appear in the query result (for now, see Issue 124).
+			if !cls.Optional {
+				return nil, &skippableError{"cls.OTypeAlias in non-optional clause", err}
+			}
+			c = &table.Cell{}
+		} else {
+			c = &table.Cell{S: table.CellString(n.Type().String())}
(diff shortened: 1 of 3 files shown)


thiagovas (approved):

Tati1701 (approved):
Thanks, Roger! LGTM :)

PR #132, opened 2020-08-13
Allow single AT extraction just after object binding in grammar

rogerlucena:
Allow the user to do a single AT extraction just after an object binding (useful when the object is a predicate, in a reification scenario for example). To illustrate: 

```
SELECT ?p, ?o, ?o_time
FROM ?test
WHERE {
    ?s ?p ?o AT ?o_time
};
```

### DIFF
--- a/bql/grammar/grammar_test.go
+++ b/bql/grammar/grammar_test.go
@@ -51,6 +51,7 @@ func TestAcceptByParse(t *testing.T) {
 		`select ?a from ?b where{?s ?p ?o as ?x id ?y type ?z};`,
 		`select ?a from ?b where{?s ?p ?o as ?x type ?y id ?z at ?t};`,
 		`select ?a from ?b where{?s ?p ?o as ?x id ?y type ?z at ?t};`,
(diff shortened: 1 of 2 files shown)


bql/grammar/grammar_test.go
```diff
@@ -51,6 +51,7 @@ func TestAcceptByParse(t *testing.T) {
 		`select ?a from ?b where{?s ?p ?o as ?x id ?y type ?z};`,
 		`select ?a from ?b where{?s ?p ?o as ?x type ?y id ?z at ?t};`,
 		`select ?a from ?b where{?s ?p ?o as ?x id ?y type ?z at ?t};`,
+		`select ?a from ?b where{?s ?p ?o at ?x};`,
```
rbkloss, line 54: Could you add a test to the planner as well?

rogerlucena: Sure. It was added with the [PR 135](https://github.com/google/badwolf/pull/135) that requires this kind of `AT` binding to work. 😀

PR #133, opened 2020-08-14
Fix problem with short variable assignment in "data_access.go"

rogerlucena:
Inside `addTriples` in "data_access.go", the short assignment `:=` was recreating a local version of the variable `err`. Then, the value eventually assigned to it was lost as soon as the `for` loop was finished (instead of keeping this value assigned to the `addTriples` scope level `err` variable). This PR comes to fix this problem.

### DIFF
--- a/bql/planner/data_access.go
+++ b/bql/planner/data_access.go
@@ -463,7 +463,8 @@ func shouldIgnoreTriple(t *triple.Triple, cls *semantic.GraphClause) (bool, erro
 func addTriples(ts <-chan *triple.Triple, cls *semantic.GraphClause, tbl *table.Table) error {
 	var err error
 	for t := range ts {
-		ignoreTriple, err := shouldIgnoreTriple(t, cls)
+		var ignoreTriple bool
+		ignoreTriple, err = shouldIgnoreTriple(t, cls)
 		if err != nil {
 			break
 		}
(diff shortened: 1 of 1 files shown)


bql/planner/data_access.go
```diff
@@ -463,15 +463,17 @@ func shouldIgnoreTriple(t *triple.Triple, cls *semantic.GraphClause) (bool, erro
 func addTriples(ts <-chan *triple.Triple, cls *semantic.GraphClause, tbl *table.Table) error {
 	var err error
 	for t := range ts {
-		ignoreTriple, err := shouldIgnoreTriple(t, cls)
+		var ignoreTriple bool
+		ignoreTriple, err = shouldIgnoreTriple(t, cls)
```
rbkloss, line 467: I ran this in go playground:
`package main

import (
	"fmt"
)

func main() {
	fmt.Println("Hello, playground")

	var i int
	for j := 0; j < 10; j++{
		i, k := j, j
		fmt.Printf("i: %d k: %d\n", i, k)
	}
	fmt.Printf("i: %d", i)
}
`
which printed :
i: 0 k: 0
i: 1 k: 1
i: 2 k: 2
i: 3 k: 3
i: 4 k: 4
i: 5 k: 5
i: 6 k: 6
i: 7 k: 7
i: 8 k: 8
i: 9 k: 9
i: 0

So the error variable should be keeping the last value

rbkloss: I misread, this is the exactly behavior you want to avoid.

rogerlucena: Yes. In the case you let above, the final value printed for `i` was `0` because this is the standard empty value for an `int` in Go.

PR #134, opened 2020-08-17
Support time binding comparisons inside HAVING clauses

rogerlucena:
If we extract a time binding inside the `WHERE` clause it would be useful to be able to use it for comparisons inside the `HAVING` clause, creating conditions around that binding. This PR comes to add support for this.

To illustrate, the query below is now possible:

```
SELECT ?p, ?time
FROM ?test
WHERE {
    /u<peter> ?p AT ?time ?o
}
HAVING ?time < 2016-03-01T00:00:00-08:00;
```

On which the time binding `?time` extracted with the `AT` keyword can now be compared to a time literal (`2016-03-01T00:00:00-08:00` above) inside the HAVING clause.

You can also combine it with `ID` bindings to create more complex conditions inside the `HAVING` clause:

```
SELECT ?s, ?p, ?p_id, ?time
FROM ?test
WHERE {
    ?s ?p ID ?p_id AT ?time ?o
}
HAVING (?p_id < "in"^^type:text) AND (?time > 2016-02-01T00:00:00-08:00);
```

**NB:** The comparisons with a time literal are not lexicographical, they indeed have the semantics of a time comparison: we consider the time zone, for example (as one should expect). 

So, if you use a condition like `?time = 2012-03-10T00:00:00-08:00` in your `HAVING` clause, you may also get triples in your query result with the `?time` binding being `2012-03-10T09:00:00+01:00` (which is equivalent to the timestamp requested in the condition, being in another time zone).

For more examples on this, please refer to the new tests added in `planner_test.go` and `expression_test.go`.

### DIFF
--- a/bql/semantic/expression.go
+++ b/bql/semantic/expression.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"reflect"
 	"strings"
+	"time"
 
 	"github.com/google/badwolf/bql/lexer"
 	"github.com/google/badwolf/bql/table"
@@ -249,6 +250,44 @@ func (e *comparisonForNodeLiteral) Evaluate(r table.Row) (bool, error) {
 	}
 }
 
+// comparisonForTimeLiteral is the internal representation of an expression of comparison between a binding and a time literal.
+type comparisonForTimeLiteral struct {
+	op OP // operation.
+
+	lB  string // left binding.
+	rTL string // right time literal.
+}
+
+func (e *comparisonForTimeLiteral) Evaluate(r table.Row) (bool, error) {
+	eL, err := cellFromRow(e.lB, r)
+	if err != nil {
+		return false, fmt.Errorf("comparisonForTimeLiteral.Evaluate failed, the call for cellFromRow(%v, %v) returned error: %v", e.lB, r, err)
+	}
+	if eL.S != nil {
+		return false, fmt.Errorf("a string binding can only be compared with a literal of type text, got literal %q instead", strings.TrimSpace(e.rTL))
+	}
+	if eL.T == nil {
+		return false, nil
+	}
+
+	timeBinding := eL.T
+	timeLiteral, err := time.Parse(time.RFC3339Nano, strings.TrimSpace(e.rTL))
+	if err != nil {
+		return false, fmt.Errorf("comparisonForTimeLiteral.Evaluate failed, could not parse time from the string %q, got error: %v", strings.TrimSpace(e.rTL), err)
+	}
+
+	switch e.op {
+	case EQ:
+		return timeBinding.Equal(timeLiteral), nil
+	case LT:
+		return timeBinding.Before(timeLiteral), nil
+	case GT:
+		return timeBinding.After(timeLiteral), nil
+	default:
+		return false, fmt.Errorf("boolean evaluation requires a boolean operation; found %q instead", e.op.String())
+	}
+}
+
 // NewEvaluationExpression creates a new evaluator for two bindings in a row.
 func NewEvaluationExpression(op OP, lB, rB string) (Evaluator, error) {
 	l, r := strings.TrimSpace(lB), strings.TrimSpace(rB)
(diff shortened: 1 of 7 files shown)


bql/semantic/expression_test.go
```diff
@@ -189,22 +190,30 @@ func TestBooleanEvaluationNode(t *testing.T) {
...
+func mustBuildNodeFromStrings(t *testing.T, nodeType, nodeID string) *node.Node {
 	n, err := node.NewNodeFromStrings(nodeType, nodeID)
 	if err != nil {
-		panic(fmt.Sprintf("could not build node from type %q and ID %q, got error: %v", nodeType, nodeID, err))
+		t.Fatalf("could not build node from type %q and ID %q, got error: %v", nodeType, nodeID, err)
 	}
 	return n
 }
 
+func mustBuildTime(t *testing.T, timeLiteral string) *time.Time {
```
jlsotomayorm, line 209: Functions that perform set up or teardown tasks such as constructing input messages are usually known as "Test helpers"

Whenever you pass a *testing.T, its' recommended to call t.Helper() to attribute failures in the test helper to the line where helper is called.

https://golang.org/pkg/testing/#B.Helper

bql/semantic/expression.go
```diff
@@ -249,6 +250,44 @@ func (e *comparisonForNodeLiteral) Evaluate(r table.Row) (bool, error) {
 	}
 }
 
+// comparisonForTimeLiteral is the internal representation of an expression of comparison between a binding and a time literal.
+type comparisonForTimeLiteral struct {
+	op OP // operation.
```
jlsotomayorm, line 255: For naming a general rule of thumb I follow and recommend is "a name should be proportional to the size of its scope and inversely proportional to the number of times that it is used within that scope"

Struct fields are usually passed around, so it's usually ok to have explicit names to understand the code without having to rely on the struct comments.

I'd recommend using: operation, leftBinding and rightTimeLiteral.  Long names should follow the Go convention of MixedCaps(https://golang.org/doc/effective_go.html#mixed-caps)

bql/semantic/expression.go
```diff
@@ -249,6 +250,44 @@ func (e *comparisonForNodeLiteral) Evaluate(r table.Row) (bool, error) {
...
+
+	lB  string // left binding.
+	rTL string // right time literal.
+}
+
+func (e *comparisonForTimeLiteral) Evaluate(r table.Row) (bool, error) {
+	eL, err := cellFromRow(e.lB, r)
```
jlsotomayorm, line 262: is this variable representing a "cell" ? maybe cL? I imagine "L" comes from left?

I'd suggest giving a little more information as this already starts getting complicated when we do eL.S.

bql/semantic/expression.go
rogerlucena: Thank you for the comments Jose. I do agree with you.
Here above I was just following the standard of the file. In this `expression.go` we have some similar structs following this convention, but I already discussed about it with Thiago and Kloss and we decided to refactor the entire file to improve readability, especially these struct attribute names. That is why I opened the PR https://github.com/google/badwolf/pull/139.
We decided to first merge the current opened PRs to master, including this 134 here, and then centralize all the refactoring in PR 139 so it will be easier to keep the same pattern for all the comparison functions.
Then, I will be doing this renaming for `rightTimeLiteral`, `leftBinding` etc there as I already started to do with the other similar structs we have there.

bql/semantic/expression.go
rogerlucena: I agree with you.
Here above I was just following the standard of the file again, but this is also going to be refactored with PR #139 for the same reason of the previous comment. I am also refactoring the correspondent variable names of the other `Evaluate` methods there.

bql/semantic/expression_test.go
rogerlucena: It is pretty nice that we have this in Go. Thank you for the comment Jose, I have just done so. 😀

bql/semantic/expression.go
jlsotomayorm: LGTM

bql/semantic/expression.go
jlsotomayorm: Got it. Thanks for explaining. Please keep in mind that for small scopes, short variables are OK. But in this case we should always try to pick name that is clear e.g 'c' for counter. If in the future refactor this will still be a short variable, we could use cell, or lcell and rcell.

PR #136, opened 2020-08-18
Use lexer.ItemTime instead of lexer.ItemPredicate for AFTER and BEFORE clauses

rogerlucena:
Previously, in `AFTER` and `BEFORE` clauses the timestamp that came after these keywords was understood as a `lexer.ItemPredicate` token in the code. 

This PR, then, comes to create a new token type: the `lexer.ItemTime`, and use it for the `AFTER` and `BEFORE` global time bound clauses, making the code cleaner.

This PR also comes to add missing tests for `AFTER` and `BEFORE` in `lexer_test.go`.


bql/lexer/lexer_test.go
```diff
@@ -255,6 +256,28 @@ func TestIndividualTokens(t *testing.T) {
 			[]Token{
 				{Type: ItemLiteral, Text: `"Hallway\"1\""^^type:text`},
 				{Type: ItemEOF}}},
+		{`AFTER 2010-03-10T00:00:00-08:00;`,
+			[]Token{
+				{Type: ItemAfter, Text: "AFTER"},
+				{Type: ItemTime, Text: `2010-03-10T00:00:00-08:00`},
+				{Type: ItemSemicolon, Text: ";"},
+				{Type: ItemEOF}}},
```
rbkloss, line 264: It is recommended for closing brackets to have the same indentation as the opening brackets, could you change accordingly, please?

rogerlucena: I agree. Here above I was just following the standard of the file while adding a new test case (as I was suggested in the past to avoid polluting a PR with "cosmetic changes").
By the end, I corrected this indentation above along with others in `lexer_test.go` isolating these improvements in one single commit. Is that ok or should I open another "cosmetic" PR?

rogerlucena:
> I recon we already had tests for this syntax in the planner, right?
> If not, do you mind adding them, please?

@rbkloss we have indeed, but I have just added two more there in `planner_test.go` to also test the case on which the `AFTER`/`BEFORE` clauses do not terminate with `;` (not being the last ones of the query, followed by `LIMIT` there).

PR #137, opened 2020-08-19
Support predicate binding equalities inside HAVING clauses

rogerlucena:
We can now compare a predicate binding to a predicate literal inside the `HAVING` clause (only equalities allowed) to filter the results of our query.

To illustrate, the following query is now possible:

```
SELECT ?s, ?p, ?o
FROM ?test
WHERE {
    ?s ?p ?o
}
HAVING ?p = "height_cm"@[];
```

Note that predicate inequalities are not allowed because their behavior may become confusing for the user (especially when time anchors come on stage). In this case an error is prompted orienting to extract bindings with the `ID`/`AT` keywords and to use these bindings to proceed with the comparisons in a clearer way.

### DIFF
--- a/bql/grammar/grammar_test.go
+++ b/bql/grammar/grammar_test.go
@@ -100,6 +100,8 @@ func TestAcceptByParse(t *testing.T) {
 		`select ?a from ?b where {?a ?p ?o} having ?b = 2014-03-10T00:00:00-08:00 limit "10"^^type:int64;`,
 		`select ?a from ?b where {?a ?p ?o} having (?b and ?b) or not (?b = ?b);`,
 		`select ?a from ?b where {?a ?p ?o} having ((?b and ?b) or not (?b = ?b));`,
+		`select ?a from ?b where {?a ?p ?o} having ?b = "foo"@[];`,
+		`select ?a from ?b where {?a ?p ?o} having ?b = "foo"@[2016-04-01T00:00:00-08:00];`,
 		// Test global time bounds.
 		`select ?a from ?b where {?s ?p ?o} before 2006-01-01T15:04:05.999999999Z07:00;`,
 		`select ?a from ?b where {?s ?p ?o} after 2006-02-03T15:04:05.999999999Z07:00;`,
--- a/bql/planner/planner_test.go
+++ b/bql/planner/planner_test.go
@@ -859,7 +859,7 @@ func TestPlannerQuery(t *testing.T) {
 		// Actual test:
 		tbl, err := plnr.Execute(ctx)
 		if err != nil {
-			t.Fatalf("planner.Execute(%s)\n= _, %v; want _, nil error", entry.q, err)
+			t.Fatalf("planner.Execute(%s)\n= _, %v; want _, nil", entry.q, err)
 		}
 		if got, want := len(tbl.Bindings()), entry.nBindings; got != want {
 			t.Errorf("planner.Execute(%s)\n= a Table with %d bindings; want %d", entry.q, got, want)
@@ -928,7 +928,7 @@ func TestPlannerQueryError(t *testing.T) {
 		// Actual test:
 		_, err = plnr.Execute(ctx)
 		if err == nil {
-			t.Errorf("planner.Execute(%s)\n= _, nil; want _, non-nil error", entry.q)
+			t.Errorf("planner.Execute(%s)\n= _, nil; want _, error", entry.q)
 		}
 	}
 }
--- a/bql/planner/planner_test.go
+++ b/bql/planner/planner_test.go
@@ -859,7 +859,7 @@ func TestPlannerQuery(t *testing.T) {
 		// Actual test:
 		tbl, err := plnr.Execute(ctx)
 		if err != nil {
-			t.Fatalf("planner.Execute(%s)\n= _, %v; want nil error", entry.q, err)
+			t.Fatalf("planner.Execute(%s)\n= _, %v; want _, nil error", entry.q, err)
 		}
 		if got, want := len(tbl.Bindings()), entry.nBindings; got != want {
 			t.Errorf("planner.Execute(%s)\n= a Table with %d bindings; want %d", entry.q, got, want)
@@ -928,7 +928,7 @@ func TestPlannerQueryError(t *testing.T) {
 		// Actual test:
 		_, err = plnr.Execute(ctx)
 		if err == nil {
-			t.Errorf("planner.Execute(%s)\n= _, nil; want non-nil error", entry.q)
+			t.Errorf("planner.Execute(%s)\n= _, nil; want _, non-nil error", entry.q)
 		}
 	}
 }
--- a/bql/planner/planner_test.go
+++ b/bql/planner/planner_test.go
@@ -807,6 +807,36 @@ func TestPlannerQuery(t *testing.T) {
 			nBindings: 2,
 			nRows:     3,
 		},
+		{
+			q: `SELECT ?s, ?p, ?o
+				FROM ?test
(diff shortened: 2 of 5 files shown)


bql/semantic/expression.go
```diff
@@ -249,6 +249,40 @@ func (e *comparisonForNodeLiteral) Evaluate(r table.Row) (bool, error) {
 	}
 }
 
+// comparisonForPredicateLiteral represents the internal representation of an expression of comparison between a binding and a predicate literal.
+type comparisonForPredicateLiteral struct {
+	op OP // operation.
+
+	lB  string // left binding.
+	rPL string // right predicate literal.
+}
```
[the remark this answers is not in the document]
rogerlucena: I agree with you. I am now opening a new PR just for isolating this refactor in one single place.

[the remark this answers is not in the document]
rogerlucena: PR refactoring `expression.go`: https://github.com/google/badwolf/pull/139
I will let it as draft till the PRs https://github.com/google/badwolf/pull/134, https://github.com/google/badwolf/pull/137 and https://github.com/google/badwolf/pull/138 are merged (as they are also making changes on this file).
But feel free to already have a look on it and, if you want, comment the changes or suggest other variable names to refactor.

bql/grammar/grammar_test.go
```diff
@@ -94,6 +94,8 @@ func TestAcceptByParse(t *testing.T) {
 		`select ?a from ?b where {?a ?p ?o} having ?b = ?b;`,
 		`select ?a from ?b where {?a ?p ?o} having (?b and ?b) or not (?b = ?b);`,
 		`select ?a from ?b where {?a ?p ?o} having ((?b and ?b) or not (?b = ?b));`,
+		`select ?a from ?b where {?a ?p ?o} having ?b = "foo"@[];`,
```
thiagovas, line 97: Can you add tests with cases that should be rejected?
At TestRejectByParse

Something like:
		`select ?a from ?b where {?a ?p ?o} having ?b < "foo"@[2016-04-01T00:00:00-08:00];`
		`select ?a from ?b where {?a ?p ?o} having ?b < "foo"@[];`
		`select ?a from ?b where {?a ?p ?o} having ?b > "foo"@[2016-04-01T00:00:00-08:00];`
		`select ?a from ?b where {?a ?p ?o} having ?b > "foo"@[];`

And any other tests you may think of.

rogerlucena: Because of the way the grammar is structured inside `havingClauses()` and `havingClausesBinaryCompositeClauses()`, we cannot differentiate a `lexer.ItemPredicate` from a `lexer.ItemLiteral`, for example, with regards to which operations are allowed (`EQ`, `GT` and `LT`) for each of them at the grammar level.

These tests, then, are indeed accepted by parse, which is good because otherwise the error message for the user would be a parsing error message instead of the instructing one I prepared for her in the case of predicate inequalities inside `expression.go` (that instruct the user to use the `ID` and `AT` keywords to do binding extractions and use them for comparisons):
 https://github.com/google/badwolf/blob/bd3e87062c91ae981519f03c21c00e96439d4be6/bql/semantic/expression.go#L282

This way, if I were to add these tests you suggested it should be to `TestAcceptByParse` instead. What do you think?

bql/planner/planner_test.go
```diff
@@ -778,7 +808,7 @@ func TestPlannerQuery(t *testing.T) {
 		// Actual test:
 		tbl, err := plnr.Execute(ctx)
 		if err != nil {
-			t.Fatalf("planner.Execute(%s)\n= _, %v; want nil error", entry.q, err)
+			t.Fatalf("planner.Execute(%s)\n= _, %v; want _, nil error", entry.q, err)
```
Tati1701, line 811: `planner.Execute(%s)\n= _, %v; want _, nil` is enough here - no need to add the `error`, as you already specify it's the second input you're comparing. It doesn't really matter what the type of the second return is, just that you got something different from what you expected.

Same for the other messages below. :)

bql/planner/planner_test.go
```diff
@@ -831,7 +877,7 @@ func TestPlannerQueryError(t *testing.T) {
 		// Actual test:
 		_, err = plnr.Execute(ctx)
 		if err == nil {
-			t.Errorf("planner.Execute(%s)\n= _, nil; want non-nil error", entry.q)
+			t.Errorf("planner.Execute(%s)\n= _, nil; want _, non-nil error", entry.q)
```
Tati1701, line 880: nit: If you had some error (even if it is a partial message or status) to add to the test cases here it would be better. Given you don't, I guess `error` should be enough (rather than `non-nil error`).

bql/semantic/expression.go
```diff
@@ -249,6 +249,40 @@ func (e *comparisonForNodeLiteral) Evaluate(r table.Row) (bool, error) {
...
+
+func (e *comparisonForPredicateLiteral) Evaluate(r table.Row) (bool, error) {
+	eL, err := cellFromRow(e.lB, r)
```
Tati1701, line 261: nit/optional: `el` is ok here, you don't need to capitalize the `L` just because it is short for literal xD

bql/semantic/expression.go
```diff
@@ -249,6 +249,40 @@ func (e *comparisonForNodeLiteral) Evaluate(r table.Row) (bool, error) {
...
+
+	csEL, err := formatCell(eL)
+	if err != nil {
+		return false, fmt.Errorf("comparisonForPredicateLiteral.Evaluate failed, the call for formatCell(%s) returned error: %v", eL, err)
+	}
+	csER := strings.TrimSpace(e.rPL)
+
+	switch e.op {
+	case EQ:
+		return reflect.DeepEqual(csEL, csER), nil
```
Tati1701, line 280: If both are strings, you can just compare them with `==`. No need for the reflect.DeepEqual.

bql/semantic/expression_test.go
```diff
@@ -715,7 +824,7 @@ func TestEvaluatorEvaluateError(t *testing.T) {
 
 		got, err := eval.Evaluate(entry.r)
 		if err == nil {
-			t.Errorf("%s: eval.Evaluate(%v) = _, nil; want non-nil error", entry.id, entry.r)
+			t.Errorf("%s: eval.Evaluate(%v) = _, nil; want _, non-nil error", entry.id, entry.r)
 		}
 		if want := entry.want; got != want {
```
Tati1701, line 829: `want` is always false for `TestEvaluatorEvaluateError` test cases right? You could remove it from the test struct and simply compare got with `false`.

```
if want := false; got != want { ... }
```

bql/planner/planner_test.go
rogerlucena: Done. Thank you Tati. =]

bql/planner/planner_test.go
rogerlucena: Done.

bql/semantic/expression.go
rogerlucena: Good to know, I thought Go was more inflexible with regards to camelCase in this scenario.
I am refactoring this `expression.go` file in the PR 139, mainly to improve some variable names, so for now I will just let it as it is as it will already be changed there.

bql/semantic/expression.go
rogerlucena: Thank you for getting this Tati. I have just done so.

bql/semantic/expression_test.go
rogerlucena: Thank you once again for pointing this Tati, I have just done so. 🙂

Issue #129, opened 2020-08-07
FILTER keyword


rogerlucena:
It would be very useful to have in BQL a `FILTER` keyword that could allow us to filter out part of the results of a query in a level closer to the storage (closer to the driver), improving performance.

Some of the functionalities that this `FILTER` keyword could accept:

1) Filtering only the triples with **immutable predicates** for our query result.

The syntax for that could be something like:

```
FILTER isImmutable(?p)
```

Coming inside or after the `WHERE` clause to specify that the predicate bound to `?p` in our query should be immutable.

Another function `isTemporal` could work similarly.

This could come as a solution for what was asked in the Issue #115. 

2) Filtering only the triples with the **latest time anchor** for our query result.

It is pretty common to be interested only in the latest triple of a time series. Instead of getting all the triples, sorting them by the time anchor in decreasing order and limiting the result to 1, as one may be doing nowadays, we could just use `FILTER` to do that in a much less expensive way using a syntax like:

```
FILTER latest(?p)
```

This is a pretty common use case already highlighted by the Issues #86 and #85. 

It also opens the possibility for supporting the opposite: filtering only the earliest time anchor, as illustrated below.

```
FILTER earliest(?p)
```

One could also decide for a syntax that uses directly the time binding for filtering, like in:

```
FILTER latest(?date)
``` 

3) Allow using **regular expressions** for matching.

We could use regex for filtering too. The syntax could be something like:

```
FILTER match(?obj, "ab+"^^type:text)
```

Which also resonates with what was asked by the Issue #122.

4) Filtering to satisfy **comparisons** (evaluated as boolean conditions).

For example:

```
FILTER greaterThan(?obj, "37"^^type:int64)
```

With the functions `lowerThan` and `equal` it should be analogous.

5) Filtering to satisfy a **combination of functions**.

For example, one could write something like:

```
FILTER latest(lowerThan(?date, 2005-01-02T15:04:05.999999999Z07:00))
```

To get in the query result only the latest element of a time series while also restricting the time interval to be before a given date.  

Another approach for this would be building a function like:

```
FILTER latestBeforeUpperBound(?date, 2005-01-02T15:04:05.999999999Z07:00)
```

6) Others.

Other ideas for filtering functions could be the likes of:

```
FILTER isToday(?date)
```

That would compare a binding with a value extracted during runtime (the current day in this example).

[...]

Issue #86, opened 2018-12-06
Allow specifying only last temporal triple on queries


xllora:
There is a common use case were temporal triples are use to maintain time series. However, common queries just require to get the last triple in the time series, not all the anchored versions. The goal of this issue is to track the modifications required to do so.

xllora:
The initial work only focus on the modifications required on the storage layer to make it possible.

tucnak:
Hello! Is there a timeline on any of these issues? BQL looks great, and kind of natural if you are coming from some SQL/SPARQL place, but the killer feature that is anchoring seems off-the-mile just enough to throw the whole thing off.

One case which I would love to make for something like BQL would be to track user actions, so easy analytics over of the whole user base experience becomes possible, but then again you would expect more flexibility in how requests are made. For certain queries you would need to specify a timeframe, and maybe some limits (sampling?) and ideally you would want to group different anchor points by their predicate in the output, so it's then easier to deal with it in the application.

I believe I could be missing something, but from what I've read, seen in the code, and tried online—it doesn't seem possible to be doing any of this at all. Unfortunately, it's not clear how to do this in cayley either. If only the long-awaited crossover episode did happen with at least some of this functionality in place, it would instantly make efficient temporal modeling over arbitrary backends not only possible, but quite easy actually. (postgres!! I believe it's quite good at nosql as well as hstore?)

P.S. It's a great pity grammatically that BQL doesn't support a trailing dot `.` for queries with multiple clauses.

xllora:
Hi @tucnak! Thanks for sharing your thoughts. 

The changes to get `latest` are in place for temporal predicates. The part that we have not decide is how to expose it in to BQL. The naive way would go with something along the lines of 

```
SELECT ?p, ?o
FROM  ?some_graph
WHERE {
    /some_type<some_id> "foo"@[LATEST]  as ?p ?o
}
```

This would allow to introduce other semantics (e.g. `FIRST`) or primitives that require extra info like your point on sampling (e.g. `SAMPLE(10)`).  These would require also some costraining on what you can pass on time bounded queries.

```
SELECT ?p, ?o
FROM  ?some_graph
WHERE {
    /some_type<some_id> "foo"@[,]  AS ?p ?o
}
```

Would be equivalent to 

```
SELECT ?p, ?o
FROM  ?some_graph
WHERE {
    /some_type<some_id> "foo"@[FIRST,LATEST]  AS ?p ?o
}
```

This could also allow some advance, more obscure querying like


```
[...]

PR #149, opened 2020-10-05
FILTER keyword for latest anchor queries (grammar/lexer/hooks)

rogerlucena:
This PR comes as the first step on #129, starting to set up a `FILTER` keyword in BadWolf at the grammar, lexer and hooks levels. The first `FILTER` function chosen to be implemented is the `latest` one, to solve what was requested by #86.

It would be very useful to have in BQL a `FILTER` keyword that could allow us to filter out part of the results of a query in a level closer to the storage (closer to the driver), improving performance. This is exactly what this PR starts to introduce, setting up the foundations for this keyword in the grammar, lexer and hooks.

The final objective is to allow the user to specify, inside of `WHERE`, which bindings they want to apply a `FILTER` to, proceeding with a more fine-grained lookup on storage, avoiding unnecessary retrieval of data and optimizing query performance.

To illustrate, queries such as the one below will in the future be possible:

```
SELECT ?s, ?p, ?o
FROM ?test
WHERE {
    ?s ?p ?o .
    FILTER latest(?p)
};
```

That would return all the temporal triples of the `?test` graph that have the latest timestamp of the time series they are part of (a recorrent use case in BadWolf), skipping immutable triples found along the way. 

Regarding their position inside `WHERE`, `FILTER` clauses must come after all the graph pattern clauses, just by the end of the `WHERE` (closer to its closing bracket). Regarding trailing dots, a `FILTER` clause is understood just like any other graph clause inside of `WHERE`: the trailing dot is mandatory at the end of each clause (`FILTER` or graph ones indistinguishably), with the exception of the last one (for which the dot is optional).

### DIFF
--- a/bql/grammar/grammar.go
+++ b/bql/grammar/grammar.go
@@ -327,6 +327,7 @@ func moreClauses() []*Clause {
 			Elements: []Element{
 				NewTokenType(lexer.ItemDot),
 				NewSymbol("CLAUSES"),
+				NewSymbol("FILTER_CLAUSES"),
 			},
 		},
 		{},
@@ -359,6 +360,33 @@ func clauses() []*Clause {
 				NewSymbol("PREDICATE"),
 				NewSymbol("OBJECT"),
 				NewSymbol("MORE_CLAUSES"),
+			},
+		},
+		{},
+	}
+}
+
+func moreFilterClauses() []*Clause {
+	return []*Clause{
+		{
+			Elements: []Element{
+				NewTokenType(lexer.ItemDot),
+				NewSymbol("FILTER_CLAUSES"),
+			},
+		},
+		{},
+	}
+}
+func filterClauses() []*Clause {
+	return []*Clause{
+		{
+			Elements: []Element{
+				NewTokenType(lexer.ItemFilter),
+				NewTokenType(lexer.ItemFilterFunction),
+				NewTokenType(lexer.ItemLPar),
+				NewTokenType(lexer.ItemBinding),
+				NewTokenType(lexer.ItemRPar),
+				NewSymbol("MORE_FILTER_CLAUSES"),
 			},
 		},
 		{},
@@ -1266,6 +1294,8 @@ func BQL() *Grammar {
 		"MORE_CLAUSES":                           moreClauses(),
 		"CLAUSES":                                clauses(),
 		"OPTIONAL_CLAUSE":                        optionalClauses(),
+		"FILTER_CLAUSES":                         filterClauses(),
+		"MORE_FILTER_CLAUSES":                    moreFilterClauses(),
 		"SUBJECT_EXTRACT":                        subjectExtractClauses(),
 		"SUBJECT_TYPE":                           subjectTypeClauses(),
 		"SUBJECT_ID":                             subjectIDClauses(),
--- a/bql/grammar/grammar.go
+++ b/bql/grammar/grammar.go
@@ -1430,6 +1430,12 @@ func SemanticBQL() *Grammar {
 	}
 	setElementHook(semanticBQL, objSymbols, semantic.WhereObjectClauseHook(), nil)
 
+	// Filter clause hook.
+	filterSymbols := []semantic.Symbol{
+		"FILTER_CLAUSES",
+	}
+	setElementHook(semanticBQL, filterSymbols, semantic.WhereFilterClauseHook(), nil)
+
 	// Collect binding variables variables.
 	varSymbols := []semantic.Symbol{
 		"VARS", "VARS_AS", "MORE_VARS", "COUNT_DISTINCT",
--- a/bql/semantic/convert.go
+++ b/bql/semantic/convert.go
@@ -69,7 +69,7 @@ func (c ConsumedElement) Token() *lexer.Token {
 	return c.token
 }
 
-// Token returns the boxed token.
+// String returns a string representation of the ConsumedElement.
 func (c ConsumedElement) String() string {
 	return fmt.Sprintf("{isSymbol=%v, symbol=%s, token=%s}", c.isSymbol, c.symbol, c.token)
 }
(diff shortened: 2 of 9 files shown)


bql/semantic/convert.go
```diff
@@ -69,6 +69,11 @@ func (c ConsumedElement) Token() *lexer.Token {
 	return c.token
 }
 
+// Token returns the boxed token.
```
jlsotomayorm, line 72: String returns. ...

bql/semantic/hooks_test.go
```diff
@@ -179,8 +179,220 @@ func TestWhereInitClauseHook(t *testing.T) {
...
+	for i, entry := range testTable {
+		for _, ce := range entry.ces {
+			if _, err := f(st, ce); err != nil {
+				t.Errorf("%q: semantic.WhereFilterClauseHook(%s) = _, %v; want _, nil", entry.id, ce, err)
+			}
+		}
+		if got, want := st.FilterClauses()[i], entry.want; !reflect.DeepEqual(got, want) {
+			t.Errorf("%q: semantic.Statement.FilterClauses()[%d] = %s; want %s", entry.id, i, got, want)
+		}
+	}
```
jlsotomayorm, line 270: I recommend using t.Run to run subtests one for each table entry. Reference:https://gobyexample.com/testing.

You can always introduce a "description" field to your table entries, and pass it as the first parameter of t.Run. This way the code is self documented because each test case has a descriptive string. e.g
       desc: "FilterClauseHook returns error"
        id: "FILTER latest(?p)",
       ces: []ConsumedElement{}

bql/semantic/hooks_test.go
```diff
@@ -179,8 +179,220 @@ func TestWhereInitClauseHook(t *testing.T) {
...
+				}),
+				NewConsumedToken(&lexer.Token{
+					Type: lexer.ItemRPar,
+					Text: ")",
+				}),
+			},
+			ceIndexError: 5,
+		},
+	}
+	for _, entry := range testTable {
```
jlsotomayorm, line 378: same here, let's use t.Run

bql/semantic/semantic_test.go
```diff
@@ -259,6 +259,32 @@ func TestGraphClauseManipulation(t *testing.T) {
 	}
 }
 
+func TestFilterClauseManipulation(t *testing.T) {
+	st := &Statement{}
+
+	t.Run("test workingFilter initial states", func(t *testing.T) {
```
jlsotomayorm, line 265: this will be printed as:

TestFilterClauseManipulation/test_workingFilter_initial_states

so we can skip the word 'test'. Also a recommendation is to include the expected return for your test case like:
"workingFilter with initial states success" / or returns OK

bql/semantic/semantic_test.go
```diff
@@ -259,6 +259,32 @@ func TestGraphClauseManipulation(t *testing.T) {
...
+		if wf := st.WorkingFilter(); wf != nil {
+			t.Fatalf(`semantic.Statement.WorkingFilter() = %q for statement "%v" without initialization; want nil`, wf, st)
+		}
+		st.ResetWorkingFilterClause()
+		if wf := st.WorkingFilter(); wf == nil || !wf.IsEmpty() {
+			t.Fatalf(`semantic.Statement.WorkingFilter() = %q for statement "%v" after call to ResetWorkingFilterClause; want empty FilterClause`, wf, st)
+		}
+	})
+
+	t.Run("test call to add workingFilter", func(t *testing.T) {
```
jlsotomayorm, line 275: same as above

bql/grammar/grammar.go
```diff
@@ -365,6 +366,33 @@ func clauses() []*Clause {
 	}
 }
 
+func moreFilterClauses() []*Clause {
```
jlsotomayorm, line 369: it looks like "more filter clauses" is part of the business logic (grammar in this case). I'm wondering if it make sense to be more specific as "more filters" is too generic. A common question would be more than what?. How much more filters? What's the difference from these "more filters" and the normal "filters".

bql/semantic/convert.go
rogerlucena: I am not sure how I let it pass. Thank you for pointing out Jose.

bql/semantic/hooks_test.go
rogerlucena: Done. Thank you Jose.

Here I used the `entry.id` as the test description since, given the name of the function, we already know that we are testing for success. For `TestWhereFilterClauseHookError` below I do the same thing since, given the name of the function, we already know that we are testing for error.

bql/semantic/hooks_test.go
rogerlucena: Done. Commented in the previous comment.

bql/semantic/semantic_test.go
rogerlucena: Done. Thank you for the tips Jose. =]

bql/semantic/semantic_test.go
rogerlucena: Done. 👍

bql/grammar/grammar.go
rogerlucena: I see what you mean. The BadWolf grammar is a left LL(1) factorized grammar as commented in the beginning of the `grammar.go` file, which means that it is defined recursively looking one step ahead each time during the parsing. 

This way, the `moreFilterClauses` comes just to recursively open a path to more filter clauses in the grammar, during parsing, as it is common in this type of grammar and as it is already done for other functions in this file (such as `moreClauses` and `moreGraphsClauses`).

Since this is intrinsically related to the way this kind of recursive grammar works, and we already have other examples like this in the file, I am inclined to let it as it is. But ltm if you still think we should make any other changes here. : )

thiagovas (approved):

bql/grammar/grammar.go
jlsotomayorm: LGTM

jlsotomayorm (approved):

bql/semantic/hooks_test.go
rogerlucena: Related Issue: https://github.com/google/badwolf/issues/152.

PR #150, opened 2020-10-05
FILTER keyword for latest anchor queries (planner/memory)

rogerlucena:
This PR comes as a part of #129, finishing to set up a `FILTER` keyword in BadWolf at the planner and memory levels. The first `FILTER` function chosen to be implemented is the `latest` one, solving what was requested by #86.

It would be very useful to have in BQL a `FILTER` keyword that could allow us to filter out part of the results of a query in a level closer to the storage (closer to the driver), improving performance. This is exactly what this PR finishes introducing, completing what started with the PR #149. This PR also showcases the full implementation of how this keyword could work on the driver/storage side as well (exemplified with the changes added to the volatile open-source driver in `memory.go` below).

Then, now the user can specify, inside of `WHERE`, which bindings they want to apply a `FILTER` to, proceeding with a more fine-grained lookup on storage, avoiding unnecessary retrieval of data and optimizing query performance.

To illustrate, queries such as the one below are now possible:

```
SELECT ?s, ?p, ?o
FROM ?test
WHERE {
    ?s ?p ?o .
    FILTER latest(?p)
};
```

That would return all the temporal triples of the `?test` graph that have the latest timestamp of the time series they are part of (a recorrent use case in BadWolf), skipping immutable triples found along the way. This FILTER function also works for objects `?o` in the case of reification/blank nodes: in this case the returned triples would be the ones on which the object is necessarily a temporal predicate with latest timestamp among the predicates with that same predicate ID (in the "object" position of the clause), analogously to what happened with `?p`.

This `FILTER` for `latest` anchor above also works for alias bindings obtained with the `AS` keyword for predicates and objects too.

At the moment only one `FILTER` is supported for each graph clause inside of `WHERE`, and only one `FILTER` is supported for each given binding as well.

In the future, to add a new `FILTER` function, the steps to follow are:

1. Add a new enum item in the list of supported filter Operations in `filter.go`;
2. Add a new entry in the `SupportedOperations` map in `filter.go` to map the **lowercase** string of the `FILTER` function being added to its correspondent `filter.Operation` element;
3. Update the `String` method of `Operation` in `filter.go`;
4. Add a new switch case inside `compatibleBindingsInClauseForFilterOperation` in `planner.go` to specify for which fields and bindings of a clause the newly added `filter.Operation` can be applied to;
5. Implement the appropriate behavior on the driver side.

### DIFF
--- a/bql/planner/filter/filter.go
+++ b/bql/planner/filter/filter.go
@@ -27,6 +27,16 @@ const (
 	Latest Operation = iota + 1
 )
 
+// Field represents the position of the semantic.GraphClause that will be operated by the filter at storage level.
+type Field int
+
+// List of filter fields.
+const (
+	SubjectField Field = iota + 1
+	PredicateField
+	ObjectField
+)
+
 // SupportedOperations maps suported filter operation strings to their correspondant Operation.
 // Note that the string keys here must be in lowercase letters only (for compatibility with the WhereFilterClauseHook).
 var SupportedOperations = map[string]Operation{
@@ -39,7 +49,7 @@ var SupportedOperations = map[string]Operation{
 // function (not applicable for all Operations - some like "latest" do not use it while others like "greaterThan" do, see Issue 129).
 type StorageOptions struct {
 	Operation Operation
-	Field     string
+	Field     Field
 	Value     string
 }
 
@@ -50,6 +60,20 @@ func (op Operation) String() string {
 		return "latest"
 	default:
 		return fmt.Sprintf(`not defined filter operation "%d"`, op)
+	}
+}
+
+// String returns the string representation of Field.
+func (f Field) String() string {
+	switch f {
+	case SubjectField:
+		return "subject field"
+	case PredicateField:
+		return "predicate field"
+	case ObjectField:
+		return "object field"
+	default:
+		return fmt.Sprintf(`not defined filter field "%d"`, f)
 	}
 }
 
--- a/bql/planner/filter/filter.go
+++ b/bql/planner/filter/filter.go
@@ -33,6 +33,16 @@ var SupportedOperations = map[string]Operation{
 	"latest": Latest,
 }
 
+// StorageOptions represent the storage level specifications for the filtering to be executed.
+// Operation below refers to the filter function being applied (eg: "latest"), Field refers to the position of the graph clause it
+// will be applied to ("subject", "predicate" or "object") and Value, when specified, contains the second argument of the filter
+// function (not applicable for all Operations - some like "latest" do not use it while others like "greaterThan" do, see Issue 129).
+type StorageOptions struct {
+	Operation Operation
+	Field     string
+	Value     string
+}
+
 // String returns the string representation of Operation.
 func (op Operation) String() string {
 	switch op {
--- a/bql/planner/filter/filter.go
+++ b/bql/planner/filter/filter.go
@@ -44,9 +44,9 @@ var SupportedOperations = map[string]Operation{
 }
 
 // StorageOptions represent the storage level specifications for the filtering to be executed.
-// Operation below refers to the filter function being applied (eg: "latest"), Field refers to the position of the graph clause it
-// will be applied to ("subject", "predicate" or "object") and Value, when specified, contains the second argument of the filter
-// function (not applicable for all Operations - some like "latest" do not use it while others like "greaterThan" do, see Issue 129).
+// Operation below refers to the filter function being applied (eg: Latest), Field refers to the position of the graph clause it
+// will be applied to (subject, predicate, or object) and Value, when specified, contains the second argument of the filter
+// function (not applicable for all Operations - some like Latest do not use it while others like GreaterThan do, see Issue 129).
 type StorageOptions struct {
 	Operation Operation
 	Field     Field
--- a/bql/planner/filter/filter.go
+++ b/bql/planner/filter/filter.go
@@ -63,6 +63,11 @@ func (op Operation) String() string {
 	}
 }
 
+// IsEmpty returns true if the Operation was not set yet.
+func (op Operation) IsEmpty() bool {
+	return op == Operation(0)
+}
+
 // String returns the string representation of Field.
 func (f Field) String() string {
 	switch f {
--- a/bql/planner/filter/filter.go
+++ b/bql/planner/filter/filter.go
@@ -1,0 +1,44 @@
+// Copyright 2020 Google Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package filter isolates core FILTER related implementation.
+package filter
+
+import (
+	"fmt"
+)
+
+// Operation represents a filter operation supported in BadWolf.
+type Operation int
+
+// List of supported filter operations.
+const (
+	Latest Operation = iota + 1
+)
+
+// SupportedOperations maps suported filter operation strings to their correspondant Operation.
+// Note that the string keys here must be in lowercase letters only (for compatibility with the WhereFilterClauseHook).
+var SupportedOperations = map[string]Operation{
+	"latest": Latest,
+}
+
+// String returns the string representation of Operation.
+func (op Operation) String() string {
+	switch op {
+	case Latest:
+		return "latest"
+	default:
+		return fmt.Sprintf(`not defined filter operation "%d"`, op)
+	}
+}
--- a/bql/planner/filter/filter.go
+++ b/bql/planner/filter/filter.go
@@ -52,3 +52,8 @@ func (op Operation) String() string {
 		return fmt.Sprintf(`not defined filter operation "%d"`, op)
 	}
 }
+
+// String returns the string representation of StorageOptions.
+func (so *StorageOptions) String() string {
+	return fmt.Sprintf("%+v", *so)
+}
--- a/bql/planner/planner.go
+++ b/bql/planner/planner.go
@@ -654,22 +654,22 @@ func organizeClausesByBinding(clauses []*semantic.GraphClause) map[string][]*sem
 	return clausesByBinding
 }
 
-// compatibleBindingsInClauseForFilterOperation returns the bindings in the given graph clause that
-// are compatible with the specified filter operation.
-func compatibleBindingsInClauseForFilterOperation(cls *semantic.GraphClause, operation string) (bindingsByField map[string]map[string]bool, err error) {
-	supportedFiltersAndBindings := map[string]map[string]map[string]bool{
-		"latest": {
-			"predicate": {cls.PBinding: true, cls.PAlias: true},
-			"object":    {cls.OBinding: true, cls.OAlias: true},
-		},
-	}
-
-	if _, ok := supportedFiltersAndBindings[operation]; !ok {
+// compatibleBindingsInClauseForFilterOperation returns a function that, for each given clause, returns the bindings that are
+// compatible with the specified filter operation.
+func compatibleBindingsInClauseForFilterOperation(operation string) (compatibleBindingsInClause func(cls *semantic.GraphClause) (bindingsByField map[string]map[string]bool), err error) {
+	switch operation {
+	case "latest":
+		compatibleBindingsInClause = func(cls *semantic.GraphClause) (bindingsByField map[string]map[string]bool) {
+			bindingsByField = map[string]map[string]bool{
+				"predicate": {cls.PBinding: true, cls.PAlias: true},
+				"object":    {cls.OBinding: true, cls.OAlias: true},
+			}
+			return
+		}
+	default:
 		err = fmt.Errorf("filter function %q on filter clause is not supported", operation)
-		return
 	}
 
-	bindingsByField = supportedFiltersAndBindings[operation]
 	return
 }
 
@@ -684,16 +684,16 @@ func organizeFilterOptionsByClause(filters []*semantic.FilterClause, clauses []*
 			return nil, fmt.Errorf("binding %q referenced by filter clause %q does not exist in the graph pattern", f.Binding, f)
 		}
 
+		compatibleBindingsInClause, err := compatibleBindingsInClauseForFilterOperation(f.Operation)
+		if err != nil {
+			return nil, err
+		}
 		for _, cls := range clausesByBinding[f.Binding] {
 			if _, ok := filterOptionsByClause[cls]; ok {
 				return nil, fmt.Errorf("multiple filters for the same graph clause or same binding are not supported at the moment")
 			}
 
-			compatibleBindingsByField, err := compatibleBindingsInClauseForFilterOperation(cls, f.Operation)
-			if err != nil {
-				return nil, err
-			}
-
+			compatibleBindingsByField := compatibleBindingsInClause(cls)
 			filterBindingIsCompatible := false
 			for field, bndgs := range compatibleBindingsByField {
 				if bndgs[f.Binding] {
--- a/bql/planner/planner.go
+++ b/bql/planner/planner.go
@@ -264,10 +264,9 @@ type queryPlan struct {
 	store storage.Store
 	// Prepared plan information.
 	bndgs     []string
-	bndgsMap  map[string]int
 	grfsNames []string
 	grfs      []storage.Graph
-	cls       []*semantic.GraphClause
+	clauses   []*semantic.GraphClause
 	filters   []*semantic.FilterClause
 	tbl       *table.Table
 	chanSize  int
@@ -293,9 +292,8 @@ func newQueryPlan(ctx context.Context, store storage.Store, stm *semantic.Statem
 		stm:       stm,
 		store:     store,
 		bndgs:     bs,
-		bndgsMap:  stm.BindingsMap(),
 		grfsNames: stm.InputGraphNames(),
-		cls:       stm.GraphPatternClauses(),
+		clauses:   stm.GraphPatternClauses(),
 		filters:   stm.FilterClauses(),
 		tbl:       t,
 		chanSize:  chanSize,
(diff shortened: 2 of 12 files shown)


storage/storage.go
```diff
@@ -43,8 +43,17 @@ type LookupOptions struct {
...
 	// just use the last available anchor.
 	LatestAnchor bool
+
+	// FilterOptions, if provided, represent the specifications for the filtering to be executed.
+	FilterOptions *FilteringOptions
 }
 
+// FilteringOptions represent the storage level specifications for the filtering to be executed.
+// Operation below refers to the filter function being applied (eg: "latest"), Field refers to the position of the graph clause it
+// will be applied to ("subject", "predicate" or "object") and Value, when specified, contains the second argument of the filter
```
thiagovas: super duper nit: missing oxford comma
```suggestion
// will be applied to ("subject", "predicate", or "object") and Value, when specified, contains the second argument of the filter
```

storage/memory/memory_test.go
```diff
@@ -303,6 +316,65 @@ func TestObjectsLatestAnchor(t *testing.T) {
 	}
 }
 
+func TestObjectsFilter(t *testing.T) {
+	ts, ctx := getTestTriplesFilter(t), context.Background()
+	g, _ := NewStore().NewGraph(ctx, "test")
+	if err := g.AddTriples(ctx, ts); err != nil {
+		t.Fatalf("g.AddTriples(_) failed failed to add test triples with error: %v", err)
```
thiagovas, line 324:
```suggestion
		t.Fatalf("g.AddTriples(_) failed to add test triples with error: %v", err)
```

storage/memory/memory_test.go
```diff
@@ -356,6 +428,65 @@ func TestSubjectsLatestAnchor(t *testing.T) {
 	}
 }
 
+func TestSubjectsFilter(t *testing.T) {
+	ts, ctx := getTestTriplesFilter(t), context.Background()
+	g, _ := NewStore().NewGraph(ctx, "test")
+	if err := g.AddTriples(ctx, ts); err != nil {
+		t.Fatalf("g.AddTriples(_) failed failed to add test triples with error: %v", err)
```
thiagovas, line 436:
```suggestion
		t.Fatalf("g.AddTriples(_) failed to add test triples with error: %v", err)
```

storage/memory/memory_test.go
```diff
@@ -406,6 +538,65 @@ func TestPredicatesForSubjectAndObjectLatestAnchor(t *testing.T) {
 	}
 }
 
+func TestPredicatesForSubjectAndObjectFilter(t *testing.T) {
+	ts, ctx := getTestTriplesFilter(t), context.Background()
+	g, _ := NewStore().NewGraph(ctx, "test")
+	if err := g.AddTriples(ctx, ts); err != nil {
+		t.Fatalf("g.AddTriples(_) failed failed to add test triples with error: %v", err)
```
thiagovas, line 546:
```suggestion
		t.Fatalf("g.AddTriples(_) failed to add test triples with error: %v", err)
```

storage/memory/memory_test.go
```diff
@@ -457,6 +648,55 @@ func TestPredicatesForSubjectLatestAnchor(t *testing.T) {
 	}
 }
 
+func TestPredicatesForSubjectFilter(t *testing.T) {
+	ts, ctx := getTestTriplesFilter(t), context.Background()
+	g, _ := NewStore().NewGraph(ctx, "test")
+	if err := g.AddTriples(ctx, ts); err != nil {
+		t.Fatalf("g.AddTriples(_) failed failed to add test triples with error: %v", err)
```
thiagovas, line 656:
```suggestion
		t.Fatalf("g.AddTriples(_) failed to add test triples with error: %v", err)
```

storage/memory/memory_test.go
```diff
@@ -507,6 +748,60 @@ func TestPredicatesForObjectLatestAnchor(t *testing.T) {
 	}
 }
 
+func TestPredicatesForObjectFilter(t *testing.T) {
+	ts, ctx := getTestTriplesFilter(t), context.Background()
+	g, _ := NewStore().NewGraph(ctx, "test")
+	if err := g.AddTriples(ctx, ts); err != nil {
+		t.Fatalf("g.AddTriples(_) failed failed to add test triples with error: %v", err)
```
thiagovas, line 756:
```suggestion
		t.Fatalf("g.AddTriples(_) failed to add test triples with error: %v", err)
```

storage/memory/memory_test.go
```diff
@@ -555,6 +850,55 @@ func TestTriplesForSubjectLatestAnchor(t *testing.T) {
 	}
 }
 
+func TestTriplesForSubjectFilter(t *testing.T) {
+	ts, ctx := getTestTriplesFilter(t), context.Background()
+	g, _ := NewStore().NewGraph(ctx, "test")
+	if err := g.AddTriples(ctx, ts); err != nil {
+		t.Fatalf("g.AddTriples(_) failed failed to add test triples with error: %v", err)
```
thiagovas, line 858:
```suggestion
		t.Fatalf("g.AddTriples(_) failed to add test triples with error: %v", err)
```

storage/memory/memory_test.go
```diff
@@ -602,6 +947,60 @@ func TestTriplesForPredicateLatestAnchor(t *testing.T) {
 	}
 }
 
+func TestTriplesForPredicateFilter(t *testing.T) {
+	ts, ctx := getTestTriplesFilter(t), context.Background()
+	g, _ := NewStore().NewGraph(ctx, "test")
+	if err := g.AddTriples(ctx, ts); err != nil {
+		t.Fatalf("g.AddTriples(_) failed failed to add test triples with error: %v", err)
```
thiagovas, line 955:
```suggestion
		t.Fatalf("g.AddTriples(_) failed to add test triples with error: %v", err)
```

storage/memory/memory_test.go
```diff
@@ -650,6 +1049,60 @@ func TestTriplesForObjectLatestAnchor(t *testing.T) {
 	}
 }
 
+func TestTriplesForObjectFilter(t *testing.T) {
+	ts, ctx := getTestTriplesFilter(t), context.Background()
+	g, _ := NewStore().NewGraph(ctx, "test")
+	if err := g.AddTriples(ctx, ts); err != nil {
+		t.Fatalf("g.AddTriples(_) failed failed to add test triples with error: %v", err)
```
thiagovas, line 1057:
```suggestion
		t.Fatalf("g.AddTriples(_) failed to add test triples with error: %v", err)
```

storage/memory/memory_test.go
```diff
@@ -740,6 +1194,65 @@ func TestTriplesForSubjectAndPredicateLatestAnchor(t *testing.T) {
 	}
 }
 
+func TestTriplesForSubjectAndPredicateFilter(t *testing.T) {
+	ts, ctx := getTestTriplesFilter(t), context.Background()
+	g, _ := NewStore().NewGraph(ctx, "test")
+	if err := g.AddTriples(ctx, ts); err != nil {
+		t.Fatalf("g.AddTriples(_) failed failed to add test triples with error: %v", err)
```
thiagovas, line 1202:
```suggestion
		t.Fatalf("g.AddTriples(_) failed to add test triples with error: %v", err)
```

storage/memory/memory_test.go
```diff
@@ -788,6 +1301,65 @@ func TestTriplesForPredicateAndObjectLatestAnchor(t *testing.T) {
 	}
 }
 
+func TestTriplesForPredicateAndObjectFilter(t *testing.T) {
+	ts, ctx := getTestTriplesFilter(t), context.Background()
+	g, _ := NewStore().NewGraph(ctx, "test")
+	if err := g.AddTriples(ctx, ts); err != nil {
+		t.Fatalf("g.AddTriples(_) failed failed to add test triples with error: %v", err)
```
thiagovas, line 1309:
```suggestion
		t.Fatalf("g.AddTriples(_) failed to add test triples with error: %v", err)
```

bql/planner/filter/filter.go
```diff
@@ -33,6 +33,16 @@ var SupportedOperations = map[string]Operation{
 	"latest": Latest,
 }
 
+// StorageOptions represent the storage level specifications for the filtering to be executed.
+// Operation below refers to the filter function being applied (eg: "latest"), Field refers to the position of the graph clause it
+// will be applied to ("subject", "predicate" or "object") and Value, when specified, contains the second argument of the filter
```
thiagovas:
```suggestion
// will be applied to ("subject", "predicate", or "object") and Value, when specified, contains the second argument of the filter
```

thiagovas (approved):

storage/storage.go
rogerlucena: Done, thank you Thiago. 👍

storage/memory/memory_test.go
rogerlucena: Done for this one and all the others you commented below. Thank you Thiago. 👍

bql/planner/filter/filter.go
```diff
@@ -44,9 +44,9 @@ var SupportedOperations = map[string]Operation{
 }
 
 // StorageOptions represent the storage level specifications for the filtering to be executed.
-// Operation below refers to the filter function being applied (eg: "latest"), Field refers to the position of the graph clause it
-// will be applied to ("subject", "predicate" or "object") and Value, when specified, contains the second argument of the filter
-// function (not applicable for all Operations - some like "latest" do not use it while others like "greaterThan" do, see Issue 129).
+// Operation below refers to the filter function being applied (eg: Latest), Field refers to the position of the graph clause it
+// will be applied to (subject, predicate or object) and Value, when specified, contains the second argument of the filter
```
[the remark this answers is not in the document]
rogerlucena: Done, thank you Thiago. 👍

bql/semantic/hooks.go
```diff
@@ -664,6 +671,59 @@ func whereObjectClause() ElementHook {
...
+			if currFilter == nil {
+				return nil, fmt.Errorf("could not add filter function %q to nil filter clause", tkn.Text)
+			}
+			if currFilter.Operation != "" {
+				return nil, fmt.Errorf("invalid filter function %q on filter clause since already set to %q", tkn.Text, currFilter.Operation)
+			}
+			if !supportedFilterFunctions[tkn.Text] {
+				return nil, fmt.Errorf("filter function %q on filter clause is not supported", tkn.Text)
+			}
+			currFilter.Operation = tkn.Text
```
rbkloss: WDYT of using st.WorkingFilter() instead of currFilter to be explicit that you are changing a external object instead of a local object?

bql/semantic/hooks.go
```diff
@@ -664,6 +671,59 @@ func whereObjectClause() ElementHook {
...
+		"latest": true,
+	}
+
+	hook = func(st *Statement, ce ConsumedElement) (ElementHook, error) {
+		if ce.IsSymbol() {
+			return hook, nil
+		}
+
+		tkn := ce.Token()
+		currFilter := st.WorkingFilter()
```
rbkloss, line 686: I am struggling to understand the scope of currFilter.
Is there any way to make it simpler?

bql/semantic/hooks.go
```diff
@@ -664,6 +671,59 @@ func whereObjectClause() ElementHook {
...
 
+// whereFilterClause returns an element hook that updates the working filter clause and,
+// if the filter clause is complete, populates the filters list of the statement.
+func whereFilterClause() ElementHook {
+	var hook ElementHook
+	supportedFilterFunctions := map[string]bool{
+		"latest": true,
+	}
+
+	hook = func(st *Statement, ce ConsumedElement) (ElementHook, error) {
```
rbkloss, line 680: This seems like a state machine,
 the state is given by st and ce and it modifies it and gets called again with the modified state until completion.

Would you mind explaining that?

bql/semantic/hooks.go
```diff
@@ -664,6 +671,59 @@ func whereObjectClause() ElementHook {
...
+		if ce.IsSymbol() {
+			return hook, nil
+		}
+
+		tkn := ce.Token()
+		currFilter := st.WorkingFilter()
+		switch tkn.Type {
+		case lexer.ItemFilterFunction:
+			if currFilter == nil {
+				return nil, fmt.Errorf("could not add filter function %q to nil filter clause", tkn.Text)
```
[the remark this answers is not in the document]
rogerlucena: Sure, no problem, I have just added a bit more of information to the error message.

bql/semantic/hooks.go
```diff
@@ -664,6 +671,59 @@ func whereObjectClause() ElementHook {
...
+			}
+			if currFilter.Operation != "" {
+				return nil, fmt.Errorf("invalid filter function %q on filter clause since already set to %q", tkn.Text, currFilter.Operation)
+			}
+			if !supportedFilterFunctions[tkn.Text] {
+				return nil, fmt.Errorf("filter function %q on filter clause is not supported", tkn.Text)
+			}
```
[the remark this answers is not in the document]
rogerlucena: Done.

bql/semantic/hooks.go
rogerlucena: I have just refactored this filter hook to make it clearer.

bql/semantic/hooks.go
rogerlucena: I have just refactored this filter hook to make it clearer.

bql/semantic/hooks.go
rogerlucena: I believe the comment 3 lines above for `whereFilterClause` is already explaining this:

```
// whereFilterClause returns an element hook that updates the working filter clause and,
// if the filter clause is complete, populates the filters list of the statement.
```

But, if you believe we still need to add any other comments just lmk and I will be glad to do so.

bql/semantic/semantic.go
```diff
@@ -143,6 +144,16 @@ type GraphClause struct {
 	OTemporal        bool
 }
 
+// FilterClause represents a FILTER clause inside WHERE.
+// Operation below refers to the filter function being applied (eg: "latest"), Binding refers to the binding it
+// will be applied to and Value, when specified, contains the second argument of the filter function (not applicable for all
+// Operations - some like "latest" do not use it while others like "greaterThan" do, see Issue 129).
```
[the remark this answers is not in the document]
rogerlucena: I had written a smaller comment before but then I was asked to explain a bit better these fields of `FilterClause` (for someone who does not have the same context as us to understand), and to give examples as I did above. This way, it is a bit bigger now.

Then, for clarity, I am inclined to let it as it is, but if you believe it is better to make it shorter and put these explanations somewhere else, with the same final effect on clarity for someone looking the first time at this portion of the code, just lmk.

bql/planner/planner.go
```diff
@@ -639,28 +642,120 @@ func (p *queryPlan) filterOnExistence(ctx context.Context, cls *semantic.GraphCl
...
+			clausesByBinding[b] = append(clausesByBinding[b], cls)
+		}
+	}
+
+	return clausesByBinding
+}
+
+// compatibleBindingsInClauseForFilterOperation returns a function that, for each given clause, returns the bindings that are
+// compatible with the specified filter operation.
+func compatibleBindingsInClauseForFilterOperation(operation filter.Operation) (compatibleBindingsInClause func(cls *semantic.GraphClause) (bindingsByField map[filter.Field]map[string]bool), err error) {
```
jlsotomayorm, line 660: Here "compatibleBindingsInClause" is a named return value.

Named return values are useful when your function returns two or three parameters of the same type. But using them just to avoid declaring a variable inside the function results in unnecessary API verbosity at the cost of minor implementation brevity.

Also "naked returns" are usually only acceptable only in small functions.

I'd advise instead to create a type alias for the function and also return the error as a separated parameter

storage/memory/memory.go
```diff
@@ -274,6 +275,63 @@ func (c *checker) CheckAndUpdate(p *predicate.Predicate) bool {
 	return true
 }
 
+// latestFilter executes the latest filter operation over memoryTriples following filterOptions.
+func latestFilter(memoryTriples map[string]*triple.Triple, pQuery *predicate.Predicate, filterOptions *filter.StorageOptions) (map[string]*triple.Triple, error) {
```
jlsotomayorm, line 279: Looks like 'map[string]*triple.Triple' is used in multiple places. Maybe create a type alias as well, wdyt?

storage/memory/memory.go
```diff
@@ -507,27 +579,29 @@ func (m *memory) TriplesForSubject(ctx context.Context, s *node.Node, lo *storag
...
-				}
-			}
+		if lo.FilterOptions != nil {
+			return fmt.Errorf("cannot have LatestAnchor and FilterOptions used at the same time inside lookup options")
+		}
+		lo.FilterOptions = &filter.StorageOptions{
+			Operation: filter.Latest,
+			Field:     filter.PredicateField,
+		}
+		// To guarantee that "lo.FilterOptions" will be cleaned at the driver level, since it was artificially created at the driver level for "LatestAnchor".
```
jlsotomayorm, line 596: optional: Given that this is done in multiple places it'd be useful to create a helper to clean up the filter options.
You can name it cleanFilterOptions and pass the "lo". Then code would read as:
defer cleanFilterOptions(lo)

rogerlucena: I like this idea and I would continue with it, but this is only being necessary here because this bool `LatestAnchor` in `lookupOptions` is not being deprecated yet. Then, for now, I decided to keep support on the OSS side of BadWolf for it too using a `FILTER latest predicate` behind the drapes, since they have the same behavior. But, since this bool will be deprecated soon, I tried to isolate it in just a block of code above `lo.FilterOptions != nil` so it will be easier to just delete it in the near future. Then, I am inclined to just let it as it is for now.

bql/planner/planner.go
rogerlucena: Thank you for the comment Jose. The intention here was just to document this `compatibleBindingsInClauseForFilterOperation` as best as possible, since it is a sensitive and important part of the `FILTER` implementation, using even the signature of the function for documentation as it is oriented in the link you sent above:

> "These names should be used to document the meaning of the return values."

I even started writing a new type alias for the `compatibleBindingsInClause` function but I felt that it was a bit confusing to have a separate type declared just for it, used only here, and having it named `compatibleBindingsInClauseType` or something in this direction. Also, I feel that it may be desirable for the API of `compatibleBindingsInClauseForFilterOperation` in this portion of the code to be clear and complete on what it is returning, with the function type specified there, for clarity and simplicity too.

This way, I am inclined to let the named return values here given the use-case explained above, but I agree with you regarding the naked returns. I thought before that it was a kind of guideline to use naked returns when using named return values, but I am seeing now that they do not necessarily need to come together. Then, I have just updated the function to explicitly return the values, which is also good for clarity as this function will be growing with time (as we add support for new FILTER functions on it).

If you have any other comments in this portion of the code just lmk through chat and I will be glad to make the respective changes here. Thank you once again for the comments Jose. =]

storage/memory/memory.go
rogerlucena: Thank you for the comment, Jose. I thought about this too, but given the way the implementation of this volatile driver in `memory.go` was thought and how this `map[string]*triple.Triple` is used in some places with not exactly the same meaning, I believe it may be clearer to just let it transparent instead of abstracting it through a type alias. But lmk through chat if you still have any other comments.

thiagovas (approved):

bql/planner/planner.go
jlsotomayorm: It's clearer now, LG

jlsotomayorm (approved):

bql/planner/planner.go
```diff
@@ -639,28 +642,120 @@ func (p *queryPlan) filterOnExistence(ctx context.Context, cls *semantic.GraphCl
...
+	switch operation {
+	case filter.Latest:
+		compatibleBindingsInClause = func(cls *semantic.GraphClause) (bindingsByField map[filter.Field]map[string]bool) {
```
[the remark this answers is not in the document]
rogerlucena: Done, thank you Jose. 🙂

PR #151, opened 2020-10-07
Structure tracer and make it more easily extendible

rogerlucena:
This PR comes mainly to structure the current tracer to make it lazily receive, instead of a limited slice of strings, a more flexible struct `tracer.Arguments`. This way it will be much easier in the future to extend it, adding more fields to be passed to the tracer inside the struct, without the need of changing all the other calls for `tracer.Trace` in code.

To illustrate:

```
// Old:
tracer.Trace(p.tracer, func() []string {
	return []string{fmt.Sprintf("Processing clause %d: %v", i, &cls)}
})

// Now:
tracer.Trace(p.tracer, func() *tracer.Arguments {
	return &tracer.Arguments {
		msgs: []string{fmt.Sprintf("Processing clause %d: %v", i, &cls)}
	}
})
```

It was deliberately decided to keep the lazy aspect of tracing here, keeping a lambda function as the second parameter for `tracer.Trace` above, as a way of avoiding the addition of too much overhead when tracing is not on.

### DIFF
--- a/bql/planner/tracer/trace.go
+++ b/bql/planner/tracer/trace.go
@@ -40,7 +40,7 @@ func init() {
 		for e := range c {
 			for _, msg := range e.tracerArgs().Msgs {
 				e.w.Write([]byte("["))
-				e.w.Write([]byte(e.t.Format("2006-01-02T15:04:05.999999-07:00")))
+				e.w.Write([]byte(e.t.Format(time.RFC3339Nano)))
 				e.w.Write([]byte("] "))
 				e.w.Write([]byte(msg))
 				e.w.Write([]byte("\n"))
--- a/bql/planner/tracer/trace.go
+++ b/bql/planner/tracer/trace.go
@@ -20,10 +20,15 @@ import (
 	"time"
 )
 
+// Arguments encapsulates the elements passed to the tracer.
+type Arguments struct {
+	Msgs []string
+}
+
 type event struct {
-	w    io.Writer
-	t    time.Time
-	msgs func() []string
+	w          io.Writer
+	t          time.Time
+	tracerArgs func() *Arguments
 }
 
 var c chan *event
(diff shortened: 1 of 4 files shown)


bql/planner/tracer/trace.go
```diff
@@ -33,7 +38,7 @@ func init() {
 
 	go func() {
 		for e := range c {
-			for _, msg := range e.msgs() {
+			for _, msg := range e.tracerArgs().Msgs {
 				e.w.Write([]byte("["))
 				e.w.Write([]byte(e.t.Format("2006-01-02T15:04:05.999999-07:00")))
```
rbkloss, line 43: Is there a way to format time that does not use a literal time format?
This reads odd, if a format name was being passed it would be more explicit what is happening.

rbkloss (approved):
There is no test for tracer, right?

bql/planner/tracer/trace.go
rogerlucena: I agree with you.

I have just refactored it to pass the time format name `time.RFC3339Nano`, and also tested it to verify if the timestamps printed by the tracer were indeed in the same expected format as before.

rogerlucena:
> There is no test for tracer, right?

I discussed about it with Thiago as well. For now a test for tracing would only test if the strings we are tracing are being properly passed to the channel, since the current tracer is very simple. This way, and also because the tracing API is still evolving, we decided we could wait a bit more to create a `trace_test.go` file. As soon as our tracer becomes more complex, we add more customizations (like flags and levels of verbosity) and our tracing API is more definitive we will be writing tests for it.

Nevertheless, as commented during the standup, I have tested the tracing files generated by the new and the old tracers for some queries, running some diffs to assure that the behavior is still the same after this PR.

thiagovas (approved):

Issue #115, opened 2020-07-08
Users should be able to specify if they want to include or not immutable triples on results from queries involving AFTER, BEFORE and BETWEEN


rogerlucena:
When using the keywords `AFTER`, `BEFORE` and `BETWEEN` to filter triples based on time bounds, the current default behavior for the query result is to include all the temporal triples satisfying the temporal condition **along with the immutable triples** present in the graph.

To illustrate, given the `?family` graph with the data below:

```
INSERT DATA INTO ?family {
  /u<joe> "parent_of"@[] /u<mary> .
  /u<joe> "parent_of"@[] /u<peter> .
  /u<peter> "height_cm"@[2007-01-02T15:04:05.999999999Z] "172"^^type:int64 .
  /u<joe> "height_cm"@[2007-01-02T15:04:05.999999999Z] "174"^^type:int64 .
  /u<mary> "height_cm"@[2009-01-02T15:04:05.999999999Z] "134"^^type:int64
};
```

If we execute the following query:

```
SELECT ?s, ?p, ?o
FROM ?family 
WHERE {
  ?s ?p ?o
} 
AFTER 2008-01-02T15:04:05.999999999Z;
```

We get:

```
Result:
?s	?p	?o
/u<mary>	"height_cm"@[2009-01-02T15:04:05.999999999Z]	"134"^^type:int64
/u<joe>	"parent_of"@[]	/u<peter>
/u<joe>	"parent_of"@[]	/u<mary>
```

But it should be allowed for the user to specify if she wanted to include or not immutable triples on the result.

You can use the code in the file below to reproduce the above results.

[specify_temporals.pdf](https://github.com/google/badwolf/files/4888516/specify_temporals.pdf)

PR #153, opened 2020-10-14
FILTER isImmutable

rogerlucena:
This PR comes to add support for the `isImmutable` FILTER function in BadWolf, as detailed in #129. It also comes to help resolving what was requested by #115.

To illustrate its implementation on the driver side, this PR also adds support for this FILTER in the volatile OSS driver (in `memory.go`).

Now, the user can write queries like: 

```
SELECT ?p, ?o
FROM ?test
WHERE {
    /u<peter> ?p ?o .
    FILTER isImmutable(?p)
};
```

That would return only rows for which the predicate `?p` is not temporal. For other examples of queries have a look at the newly added tests in `planner_test.go`.

This FILTER function also works for object bindings in the clause, along with predicate aliases and object aliases too.

### DIFF
--- a/bql/planner/planner_test.go
+++ b/bql/planner/planner_test.go
@@ -1107,6 +1107,46 @@ func TestPlannerQuery(t *testing.T) {
 			nBindings: 2,
 			nRows:     1,
 		},
+		{
+			q: `SELECT ?p, ?o
+				FROM ?test
+				WHERE {
+					/u<peter> ?p ?o .
+					FILTER isImmutable(?p)
+				};`,
+			nBindings: 2,
+			nRows:     2,
+		},
+		{
+			q: `SELECT ?s, ?p, ?o
+				FROM ?test
+				WHERE {
+					?s ?p ?o .
+					FILTER isImmutable(?o)
+				};`,
+			nBindings: 3,
+			nRows:     1,
+		},
+		{
+			q: `SELECT ?p_alias, ?o
+				FROM ?test
+				WHERE {
+					/u<peter> ?p AS ?p_alias ?o .
+					FILTER isImmutable(?p_alias)
+				};`,
+			nBindings: 2,
+			nRows:     2,
+		},
+		{
+			q: `SELECT ?s, ?p, ?o_alias
+				FROM ?test
+				WHERE {
+					?s ?p ?o AS ?o_alias .
+					FILTER isImmutable(?o_alias)
+				};`,
+			nBindings: 3,
+			nRows:     1,
+		},
 	}
 
 	s, ctx := memory.NewStore(), context.Background()
--- a/storage/memory/memory_test.go
+++ b/storage/memory/memory_test.go
@@ -352,6 +352,20 @@ func TestObjectsFilter(t *testing.T) {
 			p:    testutil.MustBuildPredicate(t, `"_predicate"@[]`),
 			want: map[string]int{`"meet"@[2021-04-10T04:21:00Z]`: 1},
 		},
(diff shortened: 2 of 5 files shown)


bql/planner/planner_test.go
```diff
@@ -1107,6 +1107,26 @@ func TestPlannerQuery(t *testing.T) {
...
+		{
+			q: `SELECT ?s, ?p, ?o
+				FROM ?test
+				WHERE {
+					?s ?p ?o .
+					FILTER isImmutable(?o)
+				};`,
+			nBindings: 3,
+			nRows:     1,
+		},
```
thiagovas: Can you add some tests considering aliases as well?

thiagovas (commented):
Can you add some tests in memory_test.go as well?

bql/planner/planner_test.go
rogerlucena: Sure, thank you for the comment Thiago, I have just done so. =]

rogerlucena:
> Can you add some tests in memory_test.go as well?

In the second to last commit I have already added a total of 22 new tests in `memory_test.go` too: 2 for each one of the 11 possible calls to the driver (one test for `FILTER isImmutable predicate field` and one for `FILTER isImmutable object field` for each possible call).

thiagovas (approved):

storage/memory/memory_test.go
```diff
@@ -352,6 +352,20 @@ func TestObjectsFilter(t *testing.T) {
...
+			p:    testutil.MustBuildPredicate(t, `"parent_of"@[]`),
+			want: map[string]int{"/u<paul>": 1},
+		},
+		{
+			id:   "FILTER isImmutable object",
+			lo:   &storage.LookupOptions{FilterOptions: &filter.StorageOptions{Operation: filter.IsImmutable, Field: filter.ObjectField}},
+			s:    testutil.MustBuildNodeFromStrings(t, "/_", "bn"),
+			p:    testutil.MustBuildPredicate(t, `"_predicate"@[]`),
+			want: map[string]int{`"height_cm"@[]`: 1},
+		},
```
rbkloss, line 368: Could you change the name of the test struct a bit?
You could explain what is the want, e.g., 
want -> ExpectedCountByObjectName.
s -> query_subject
p-> query_predicate

Also the logic in the test body seems a bit too complex for tests.
Instead of decrementing the map counter and deleting if the count gets to 0,
WDYT of building the "got" map and calling equal on both?

instead of:
```go
for s := range ss {
  sStr := s.String()
  if _, ok := entry.want[sStr]; !ok {

  t.Fatalf("g.Subjects(%s, %s, %s) retrieved unexpected %s", p, o, entry.lo, sStr)						 
  }
  entry.want[sStr] = entry.want[sStr] - 1
  if entry.want[sStr] == 0 {
    delete(entry.want, sStr)
  }
}
```
something like:
```go
got := map[string]int{}
for o := range os {
  oStr := s.String()
  got[oStr] += 1
}
if !cmp.Equals(entry.want, got){
  t.Fatalf("g.Subjects(%s, %s, %s) retrieved unexpected %s", p, o, entry.lo, sStr)
}
```

rbkloss (approved):

storage/memory/memory_test.go
rogerlucena: As we commented through chat, I can be doing this in a separate PR after this PR and the PR 154 are merged. Also, I will be letting this task as P2 for now to focus on current tasks, as we commented through chat as well.

PR #154, opened 2020-10-15
FILTER isTemporal

rogerlucena:
This PR comes to add support for the `isTemporal` FILTER function in BadWolf, as detailed in #129. It also comes to help resolving what was requested by #115.

To illustrate its implementation on the driver side, this PR also adds support for this FILTER in the volatile OSS driver (in `memory.go`).

Now, the user can write queries like: 

```
SELECT ?p, ?o
FROM ?test
WHERE {
    /u<peter> ?p ?o .
    FILTER isTemporal(?p)
};
```

That would return only rows for which the predicate `?p` is temporal. For other examples of queries have a look at the newly added tests in `planner_test.go`.

This FILTER function also works for object bindings in the clause, along with predicate aliases and object aliases too.

### DIFF
--- a/storage/memory/memory_test.go
+++ b/storage/memory/memory_test.go
@@ -366,6 +366,20 @@ func TestObjectsFilter(t *testing.T) {
 			p:    testutil.MustBuildPredicate(t, `"_predicate"@[]`),
 			want: map[string]int{`"height_cm"@[]`: 1},
 		},
+		{
+			id:   "FILTER isTemporal predicate",
+			lo:   &storage.LookupOptions{FilterOptions: &filter.StorageOptions{Operation: filter.IsTemporal, Field: filter.PredicateField}},
+			s:    testutil.MustBuildNodeFromStrings(t, "/u", "john"),
+			p:    testutil.MustBuildPredicate(t, `"parent_of"@[]`),
+			want: map[string]int{},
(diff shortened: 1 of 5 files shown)


thiagovas (approved):

storage/memory/memory_test.go
```diff
@@ -579,6 +607,20 @@ func TestPredicatesForSubjectAndObjectFilter(t *testing.T) {
...
+			want: map[string]int{`"meet"@[2014-04-10T04:21:00Z]`: 1},
+		},
+		{
+			id:   "FILTER isTemporal object",
+			lo:   &storage.LookupOptions{FilterOptions: &filter.StorageOptions{Operation: filter.IsTemporal, Field: filter.ObjectField}},
+			s:    testutil.MustBuildNodeFromStrings(t, "/_", "bn"),
+			o:    triple.NewPredicateObject(testutil.MustBuildPredicate(t, `"meet"@[2020-04-10T04:21:00Z]`)),
+			want: map[string]int{`"_predicate"@[]`: 1},
+		},
 	}
```
rbkloss, line 624: The same comments from the previous PR applies here.
PTAL.

rogerlucena: Commented on the PR 153.

PR #155, opened 2020-10-23
Support pprof profiling in BadWolf

rogerlucena:
This PR comes to allow the user to have more visibility on latency and memory metrics when using BadWolf, adding support for [pprof](https://github.com/google/pprof) profiling through BadWolf's CLI.

When in the terminal, after already entering the BadWolf's CLI with the command `bw bql`, the user can now write:

```
start profiling;
```

To start `pprof` profiling. The user can also specify the CPU profiling rate with the `-cpurate` flag below:

```
start profiling -cpurate <samples_per_second>
```

Similarly, to stop it they can just write:

```
stop profiling;
```

With profiling enabled, the metrics for latency and memory consumption for the BQLs executed will be printed into two files, `cpuprofile` and `memprofile`, that can then be consumed and visualized through the `go pprof` CLI tool. 

Commands such as `go tool pprof -http=":8081" bw cpuprofile` open in the browser an interface to visualize the CPU metrics: the user can see the graph of function calls made and the latencies involved, useful to identify bottlenecks; they can visualize the correspondent Flame Graph as well, and also an ordered list of the heaviest function calls made. Another useful command is `go tool pprof -pdf bw cpuprofile > cpuprofile.pdf` to generate pdfs from the profiling files. With `memprofile` the previous `pprof` commands work similarly.

### DIFF
--- a/tools/vcli/bw/repl/repl.go
+++ b/tools/vcli/bw/repl/repl.go
@@ -22,7 +22,9 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"runtime"
 	"runtime/pprof"
+	"strconv"
 	"strings"
 	"time"
 
@@ -217,13 +219,33 @@ func REPL(od storage.Store, input *os.File, rl ReadLiner, chanSize, bulkSize, bu
 			continue
 		}
 		if strings.HasPrefix(l, "start profiling") {
-			profilingEnabled = true
-			fmt.Println("Profiling with pprof is enabled.")
+			args := strings.Split(strings.TrimSpace(l)[:len(l)-1], " ")
+			switch len(args) {
+			case 2:
+				profilingEnabled = true
+				fmt.Println("Profiling with pprof is enabled.")
+			case 4:
+				if args[2] != "-cpurate" {
+					fmt.Printf("Invalid syntax with %q.\n\tstart profiling -cpurate <samples_per_second>\n", args[2])
+					break
+				}
+				cpuProfRate, err := strconv.ParseInt(args[3], 10, 32)
+				if err != nil {
+					fmt.Println(err)
+					break
+				}
+				runtime.SetCPUProfileRate(int(cpuProfRate))
+				profilingEnabled = true
+				fmt.Printf("Profiling with pprof is enabled (CPU profiling rate: %d samples per second).\n", cpuProfRate)
+			default:
+				fmt.Println("Invalid syntax.\n\tstart profiling -cpurate <samples_per_second>")
+			}
 			done <- false
 			continue
 		}
 		if strings.HasPrefix(l, "stop profiling") {
 			profilingEnabled = false
+			runtime.SetCPUProfileRate(0)
 			fmt.Println("Profiling with pprof is disabled.")
 			done <- false
 			continue
--- a/tools/vcli/bw/repl/repl.go
+++ b/tools/vcli/bw/repl/repl.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"runtime/pprof"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


thiagovas (approved):

tools/vcli/bw/repl/repl.go
```diff
@@ -175,6 +220,59 @@ func REPL(od storage.Store, input *os.File, rl ReadLiner, chanSize, bulkSize, bu
 			done <- false
 			continue
 		}
+		if strings.HasPrefix(l, "start profiling") {
```
Tati1701, line 223: For a separate PR/Unralated to your changes:
Consider updating the loop to have a switch instead of several `if strings.HasPrefix` checks. 
Remember: in Go, switch statements are more general than in other languages, so you can have cases checking for `HasPrefix`. 
https://golang.org/doc/effective_go.html#switch
Besides being more readable, it can probably simplify the `done <- false/continue` repetitions. You might want to look at labels to control the flow here, rather than using the channel :)

Tati1701 (approved):
A couple of small comments on indenting the error flow so it is more clear to the reader. 

I'll go ahead and approve but PTAL at them before submitting.

tools/vcli/bw/repl/repl.go
rogerlucena: Thank you Tati! It is cool that `switch` can be this general in Go, we will keep this in mind for a future PR as well. =]

tools/vcli/bw/repl/repl.go
```diff
@@ -175,6 +220,59 @@ func REPL(od storage.Store, input *os.File, rl ReadLiner, chanSize, bulkSize, bu
...
+			var err error
+			switch len(args) {
+			case 2:
+				if cpuProfile, memProfile, err = startProfiling(); err == nil {
+					isProfiling = true
+					fmt.Println("Profiling with pprof is on.")
+				} else {
+					fmt.Println(err)
+					fmt.Println("Profiling failed to start.")
+				}
```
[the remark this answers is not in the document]
rogerlucena: Thank you for the comment Tati! I like this suggestion to make the flow clearer, and I have just proceeded so 😄

tools/vcli/bw/repl/repl.go
```diff
@@ -175,6 +220,59 @@ func REPL(od storage.Store, input *os.File, rl ReadLiner, chanSize, bulkSize, bu
...
+					break
+				}
+				runtime.SetCPUProfileRate(int(cpuProfRate))
+				if cpuProfile, memProfile, err = startProfiling(); err == nil {
+					isProfiling = true
+					fmt.Printf("Profiling with pprof is on (CPU profiling rate: %d samples per second).\n", cpuProfRate)
+				} else {
+					fmt.Println(err)
+					fmt.Println("Profiling failed to start.")
+				}
```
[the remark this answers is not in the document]
rogerlucena: Done. =]

tools/vcli/bw/repl/repl.go
```diff
@@ -88,10 +89,50 @@ func SimpleReadLine(done chan bool) <-chan string {
...
+		return nil, nil
+	}
+	pprof.StartCPUProfile(cpuProfile)
+
+	return cpuProfile, memProfile
+}
+
+// stopProfilingIfEnabled stops profiling for the current call to runBQLFromFile if profilingEnabled is true,
+// writing the final profiling metrics to the correspondent files and closing them.
+func stopProfilingIfEnabled(profilingEnabled bool, cpuProfile, memProfile *os.File) {
```
rbkloss: Usually, it is not a good practice to receive boolean parameters because it is usually passed as a literal which is not very meaningful.
Would you mind having  a parameter struct here?
then when calling it would read stopProfilingIfEnabled(params{profilingEnabled: True...})

tools/vcli/bw/repl/repl.go
```diff
@@ -88,10 +89,50 @@ func SimpleReadLine(done chan bool) <-chan string {
 	return c
 }
 
+// startProfilingIfEnabled starts profiling for the current call to runBQLFromFile if profilingEnabled is true,
+// returning the two files to which the profiling metrics will be printed.
+func startProfilingIfEnabled(profilingEnabled bool) (*os.File, *os.File) {
```
rbkloss: I am not sure if it would be better to have a function startProfiling()
and check if it is enabled elsewhere before calling startProfiling

rbkloss (approved):

tools/vcli/bw/repl/repl.go
rogerlucena: I do not like this `IfEnabled` as well, when I first wrote this I also had mixed feelings about it. That is why in a later commit I change this to be just a `startProfiling` function, and check if it is enabled elsewhere (out of the scope of the function).

tools/vcli/bw/repl/repl.go
rogerlucena: I see your point, it is indeed more meaningful to pass something like `profilingEnabled: True` using a parameter struct. By the end, this function was changed and does not receive a boolean anymore, so the code above is outdated.

PR #156, opened 2020-10-26
Add support for FILTER Value at the grammar and hooks levels

rogerlucena:
Since there will be `FILTER` functions that require a `Value` parameter (such as `greaterThan` in the future), we must have support for a `Value` argument in the BadWolf's grammar for `FILTER` as well. But, at the same time, we must be able to differentiate these `FILTER` functions from others that do not require `Value` (such as `latest`), for a fine-grained error checking.

This PR, then, comes to implement this above, adding support for `Value` in the BadWolf's `FILTER` grammar while also checking its validity at the hooks level.

After this PR, the steps to add support for a new FILTER function become:

1. Add a new enum item in the list of supported filter Operations in `filter.go`;
2. Add a new entry in the `SupportedOperations` map in `filter.go` to map the **lowercase** string of the filter function being added to its correspondent `filter.Operation` element;
3. If this new filter function requires a `Value` parameter, add the correspondent `filter.Operation` to the `OperationRequiresValue` hash set in `filter.go`;
4. Update the `String` method of `Operation` in `filter.go`;
5. Add a new switch case inside `compatibleBindingsInClauseForFilterOperation` in `planner.go` to specify for which fields and bindings of a clause the newly added `filter.Operation` can be applied to;
6. Implement the appropriate behavior on the driver side.

### DIFF
--- a/bql/grammar/grammar_test.go
+++ b/bql/grammar/grammar_test.go
@@ -208,6 +208,12 @@ func TestAcceptByParse(t *testing.T) {
 			FILTER latest(?p) .
 			FILTER latest(?o)
 		 };`,
+		`select ?a
+		 from ?b
+		 where {
+			?s ?p ?o .
+			FILTER greaterThan(?o, "37"^^type:int64)
+		 };`,
 		// Test optional trailing dot after the last clause inside WHERE.
 		`select ?a
 		 from ?b
@@ -453,6 +459,18 @@ func TestRejectByParse(t *testing.T) {
 			?s ?p ?o .
 			FILTER late^st(?p)
 		 };`,
+		`select ?a
+		 from ?b
+		 where {
+			?s ?p ?o .
+			FILTER greaterThan(?o "37"^^type:int64)
+		 };`,
+		`select ?a
+		 from ?b
+		 where {
+			?s ?p ?o .
+			FILTER greaterThan(?o, /u<paul>)
+		 };`,
 		// Test invalid trailing dot use inside WHERE.
 		`select ?a
 		 from ?b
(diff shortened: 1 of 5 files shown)


rbkloss (approved):

bql/grammar/grammar_test.go
```diff
@@ -453,6 +459,18 @@ func TestRejectByParse(t *testing.T) {
 			?s ?p ?o .
 			FILTER late^st(?p)
 		 };`,
+		`select ?a
+		 from ?b
+		 where {
+			?s ?p ?o .
+			FILTER greaterThan(?o "37"^^type:int64)
+		 };`,
```
thiagovas, line 467: I guess I'm missing something.
How come this test is on both TestRejectByParse and TestAcceptByParse?

bql/semantic/hooks.go
```diff
@@ -706,9 +703,38 @@ func addBindingToWorkingFilter(bndg string, workingFilter *FilterClause) error {
...
+		return fmt.Errorf("filter clause Operation is missing")
+	}
+	if f.Binding == "" {
+		return fmt.Errorf("filter clause Binding is missing")
+	}
+	if f.Value == "" && filter.OperationRequiresValue[f.Operation] {
+		return fmt.Errorf("filter clause Value is required for filter Operation %q", f.Operation)
+	}
+	if f.Value != "" && !filter.OperationRequiresValue[f.Operation] {
+		return fmt.Errorf("filter clause Value is inconsistent with filter Operation %q", f.Operation)
```
thiagovas, line 734: I guess we can make this error more clear.
What about something like:
`filter clause Value is not required for filter Operation %q`

Or something in these lines?

bql/grammar/grammar_test.go
rogerlucena: In `TestRejectByParse` there is a comma missing between the binding and the value.

bql/semantic/hooks.go
rogerlucena: When writing this error message I was in doubt between "inconsistent" and "not required" too. Thank you for the comment Thiago, I have just proceeded with "not required" then. : )

bql/grammar/grammar_test.go
thiagovas: heauehaueh, I would never notice this.
Thanks!

thiagovas (approved):

jlsotomayorm (approved):

PR #157, opened 2020-11-03
Add more metadata to tracer to improve debugability

rogerlucena:
This PR comes to enhance **debugability** in BadWolf, adding more useful metadata to be traced to help in the occasion of investigating production issues.

To be more precise, this PR comes mainly to add for tracing:

- More **metadata on clause processing** in the planner level (especially the **latency for processing each clause** on the graph pattern and the **overall latency for the entire pattern** as well); 

- The **number of triples returned from driver** for each driver call, along with the **specific name of the graph** being queried during this call and also the **number of rows actually being added to the result table** after that;

- The **number of rows removed from the result table** after processing the **HAVING clause**.

### DIFF
--- a/bql/planner/data_access.go
+++ b/bql/planner/data_access.go
@@ -504,8 +504,10 @@ func addTriples(ts <-chan *triple.Triple, cls *semantic.GraphClause, tbl *table.
 	// Drain the channel to avoid leaking goroutines in the case the loop below is interrupted by an error.
 	defer drainChannel(ts)
 
+	nTrpls := 0
 	nRowsAdded := 0
 	for t := range ts {
+		nTrpls++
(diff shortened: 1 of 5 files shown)


rbkloss (approved):
Approved.
Just keep in mind the offline discussion we had regarding the memory consumption of the captured things in the tracer.

rogerlucena:
> Approved.
> Just keep in mind the offline discussion we had regarding the memory consumption of the captured things in the tracer.

Since in `trace.go` there is a separate goroutine always consuming the trace events as they arrive, working like an "automatic flush" writing to the file for each new event arriving through the channel `c`, and as discussed offline with you and Thiago, I believe we do not need to worry about this memory consumption of the captured variables anymore.

thiagovas (approved):

PR #158, opened 2020-11-06
Support verbosity levels configuration for tracing

rogerlucena:
This PR comes to support **different verbosity levels for tracing** in BadWolf, and also to allow the user to customize the current tracing verbosity through the BadWolf's CLI.

This way, it is now possible for the user to control the amount of information they want in the tracing output, accordingly to the use case, and also to control how much of computing power they want to dedicate for tracing.

In the CLI, the syntax for tracing now is:

```
start tracing [-v verbosity_level] [trace_file];
```

With the `-v` flag indicating the global verbosity level to be set for the current tracer.

At the moment, each tracing message in code can be allocated to one of the three following verbosity levels:

- **Level 1:** This level contains the **most crucial** and basic tracing messages to understand the flow of the program. Examples of messages that come here:
  - The one specifying **which query is being executed** at the moment; 
  - The one with the **final number of rows returned** after each created Plan was executed.

- **Level 2:** Tracing messages with **intermediate importance**. Examples of messages that come here:
  - The one specifying **which clause is being processed** (planner level);
  - The one specifying the **latency** of this clause processing (planner level); 
  - The one detailing the **number of triples received from the driver** after each driver call and the **number of rows** actually added to the result table after that.

- **Level 3:** The **less crucial** tracing messages, whose purpose is, mainly, to add additional details to the tracing log. Examples of messages that could come here (both examples extracted from planner):
  - “Processing **optional** clause”;
  - “Some clause bindings exist”.

If the current tracer has its global verbosity level set to `L` (after a command in the CLI), then all tracing messages with intrinsic verbosity `<= L` will be printed to the tracing output. 

Then, a global verbosity level of `1` means a very light tracer, printing to the output only what is essential, while a verbosity level of `3` means a very verbose tracer printing to the output all available tracing messages.

### DIFF
--- a/bql/planner/tracer/trace.go
+++ b/bql/planner/tracer/trace.go
@@ -31,9 +31,19 @@ type event struct {
 	tracerArgs func() *Arguments
 }
 
+// MessageTracer encapsulates the intrinsic verbosity of a given tracing message.
+type MessageTracer struct {
+	verbosity int
+}
+
+// tracerVerbosity represents the global verbosity level of the current tracer. Level 1 means minimum verbosity (printing
+// only what is crucial) while level 3 means maximum verbosity (printing all available tracing messages).
+var tracerVerbosity int
+
 var c chan *event
 
 func init() {
+	tracerVerbosity = 1          // The default tracer has minimum verbosity.
 	c = make(chan *event, 10000) // Large enought to avoid blocking as much as possible.
 
 	go func() {
@@ -49,11 +59,47 @@ func init() {
 	}()
 }
 
-// Trace attempts to write a trace if a valid writer is provided. The
-// tracer is lazy on the arguments generation to avoid adding too much
-// overhead when tracing is not on.
-func Trace(w io.Writer, tracerArgs func() *Arguments) {
-	if w == nil {
+// SetVerbosity sets the global verbosity of the current tracer to the value received as
+// input, 1 meaning minimum and 3 meaning maximum verbosity. If the received value is not
+// in the interval [1, 3], then it is truncated. The function returns the actual verbosity set.
+func SetVerbosity(verbosity int) int {
+	// Truncate verbosity if out of the range supported.
+	if verbosity < 1 {
+		verbosity = 1
+	} else if verbosity > 3 {
+		verbosity = 3
+	}
+
+	tracerVerbosity = verbosity
+	return tracerVerbosity
+}
+
+// V returns a MessageTracer with the specified verbosity level. Level 1 here means that the correspondent
+// message has the highest priority and will always be printed to the tracing output, while 3 means that this message
+// has the lowest priority and will be printed to the output only if the current tracer has maximum tracerVerbosity.
+// If the received verbosity level is out of the range [1, 3] supported, then it is truncated.
+func V(verbosity int) MessageTracer {
+	// Truncate verbosity if out of the range supported.
+	if verbosity < 1 {
+		verbosity = 1
+	} else if verbosity > 3 {
+		verbosity = 3
+	}
+
+	return MessageTracer{verbosity}
+}
+
+// isTraceable returns true if the current tracer is verbose enough to let the given MessageTracer
+// indeed trace its correspondent message.
+func (t MessageTracer) isTraceable() bool {
+	return t.verbosity <= tracerVerbosity
+}
+
+// Trace attempts to write a trace if a valid writer is provided and the verbosity level
+// of the MessageTracer is coherent with the global tracer verbosity. The tracer is lazy
+// on the arguments generation to avoid adding too much overhead when tracing is not on.
+func (t MessageTracer) Trace(w io.Writer, tracerArgs func() *Arguments) {
+	if w == nil || !t.isTraceable() {
 		return
 	}
 	c <- &event{w, time.Now(), tracerArgs}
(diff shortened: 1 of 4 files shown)


tools/vcli/bw/repl/repl.go
```diff
@@ -467,14 +467,14 @@ func planBQL(ctx context.Context, bql string, s storage.Store, chanSize, bulkSiz
 	pln, err := planner.New(ctx, s, stm, chanSize, bulkSize, w)
 	if err != nil {
 		msg := fmt.Errorf("planer.New failed failed; %v", err)
```
thiagovas, line 509: Little typo here: planner
Also failed is written twice.

thiagovas (approved):

tools/vcli/bw/repl/repl.go
rogerlucena: Done, thank you Thiago. =]

bql/planner/tracer/trace.go
```diff
@@ -25,19 +25,32 @@ type Arguments struct {
...
 
-var c chan *event
+// MessageTracer encapsulates the intrinsic verbosity of a given tracing message.
+type MessageTracer struct {
+	verbosity int
+}
+
+// tracerVerbosity represents the global verbosity level of the current tracer. Level 1 means minimum verbosity (printing
+// only what is crucial) while level 3 means maximum verbosity (printing all available tracing messages).
+var tracerVerbosity int
```
jlsotomayorm, line 42: did you consider using iota?

Iota is the "golang enum". Iota represents successive untypes integer constants which seems like what you need here PTAL: https://golang.org/ref/spec#Iota

jlsotomayorm, line 42: nit: did you consider using iota for the data type?

Iota is the "golang enum". Iota represents successive untypes integer constants which seems what you need here PTAL: https://golang.org/ref/spec#Iota

By giving a name to the integers we can also improve the readability and instead of SetVerbosity(1) do a SetVerbosity(PRINT_ONLY) or MINIMUM, wdyt?

bql/planner/tracer/trace.go
```diff
@@ -25,19 +25,32 @@ type Arguments struct {
...
+// event encapsulates a single tracing event.
 type event struct {
 	w          io.Writer
 	t          time.Time
 	tracerArgs func() *Arguments
 }
 
-var c chan *event
+// MessageTracer encapsulates the intrinsic verbosity of a given tracing message.
+type MessageTracer struct {
```
jlsotomayorm, line 36: MessageTracer doesn't seem to be used in other packages. Is this expected? If yes, maybe we should make it unexported (lowercase)

jlsotomayorm: Also I'm wondering if it's possible to avoid reading and writing a package-level variable. Does it make sense to have a constructor for the message tracer where we pass the "global verbosity"?

bql/planner/tracer/trace.go
rogerlucena: Thank you for the comment Jose. I do use `iota` in other places of the code (for the `FILTER` function types in `filter.go`, for example), but for the verbosity levels above we discussed a little about and, to follow the internal code that is often using `.V(1)` (with the explicit `int` level being passed), we decided to continue passing the explicit `int` level as well, for consistency.

Since the comments of the tracing methods are already being explicit about the semantics of these levels, it should not be a problem for a developer looking at the code to understand them. Also, the current way is a bit more concise if we compare to something like `.V(tracer.MINIMUM)` before each call to `.Trace`.

bql/planner/tracer/trace.go
rogerlucena: About `MessageTracer`, I did indeed make it unexported (lowercase) when first writing this code. But, since the public/exported `tracer.V` method was returning a `tracer.messageTracer` unexported type (to have its public `.Trace` method used in other packages), I was getting a persistent warning here on `VSCode` at the line the `tracer.V` method was being defined (saying that it was a public method return a not-public type).

Given this, to solve this warning, I decided to make `MessageTracer` a exported type.

In this situation, should I keep it exported to resolve the warning or make it unexported as I did at the beginning and ignore this warning in the IDE?

rogerlucena: About your second question, the objective of the `MessageTracer` is just to encapsulate the `.Trace` method for a given message, keeping track of the intrinsic verbosity level of that message and comparing it to the global verbosity of the current tracer before sending it to the tracing output. Its "constructor" is just the `tracer.V` method at the moment, and having the "global tracer verbosity" passed inside of it would be out of its scope.

But I understand you when worrying about "reading and writing a package-level variable". We also think about this and for the future of the tracer we are planning to make it an object, to be created by a `.New` method, having inside of it this "global tracer verbosity" field, and having the correspondent methods the `tracer` package has at the moment. This way, we would not have reads and writes to a package-level variable anymore.

jlsotomayorm: 1. Right, the warning makes sense, as you are accessing the Tracer method we should maintain as exported.

2.I agree. Probably better to have a New constructor for a generic Tracer struct and from that struct create a message tracer with a specific verbosity. Maybe leave a TODO?

jlsotomayorm (approved):

bql/planner/tracer/trace.go
rogerlucena: In the design doc we have here for the future of the tracer we already have a TODO there: an entire section dedicated for this future refactoring that will create this `.New()`, and make the tracer a struct with methods.

Thank you again Jose.

rbkloss (approved):

PR #159, opened 2020-11-07
Fix issues with tracer and improve tracing messages

rogerlucena:
This PR comes mainly to:

- Fix problem of **loop variables being passed by reference to the tracing closure** and then having only its last value being shown in the tracing output;

- Fix problem of `lookupOptions` being **passed as a pointer to the tracing closure** (in the "global lookup options" tracing message, for example, another value was being written to the output since the referenced content was changed in the meantime between the call to `.Trace` and the generation of the string message when calling the closure);

- Avoid **unnecessary and more expensive captures** inside the tracing closures (such as capturing an entire `graph` in the closure just to get its `ID` later).

### DIFF
--- a/bql/planner/data_access.go
+++ b/bql/planner/data_access.go
@@ -127,6 +127,7 @@ func simpleExist(ctx context.Context, gs []storage.Graph, cls *semantic.GraphCla
 func simpleFetch(ctx context.Context, gs []storage.Graph, cls *semantic.GraphClause, lo *storage.LookupOptions, stmLimit int64, chanSize int, w io.Writer) (*table.Table, error) {
 	s, p, o := cls.S, cls.P, cls.O
 	lo = updateTimeBounds(lo, cls)
+	loStr := lo.String()
 	tbl, err := table.New(cls.Bindings())
 	if err != nil {
 		return nil, err
@@ -171,7 +172,7 @@ func simpleFetch(ctx context.Context, gs []storage.Graph, cls *semantic.GraphCla
 			)
 			tracer.V(2).Trace(w, func() *tracer.Arguments {
 				return &tracer.Arguments{
-					Msgs: []string{fmt.Sprintf("g.Objects(%v, %v, %v), graph: %s", s, p, lo, gID)},
+					Msgs: []string{fmt.Sprintf("g.Objects(%v, %v, %s), graph: %s", s, p, loStr, gID)},
 				}
 			})
 			wg.Add(2)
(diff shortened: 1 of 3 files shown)


thiagovas (approved):

bql/planner/data_access.go
```diff
@@ -127,6 +127,7 @@ func simpleExist(ctx context.Context, gs []storage.Graph, cls *semantic.GraphCla
 func simpleFetch(ctx context.Context, gs []storage.Graph, cls *semantic.GraphClause, lo *storage.LookupOptions, stmLimit int64, chanSize int, w io.Writer) (*table.Table, error) {
 	s, p, o := cls.S, cls.P, cls.O
 	lo = updateTimeBounds(lo, cls)
+	loStr := lo.String()
```
rbkloss, line 130: I am not sure if caching lo.String to a variable really brings advantages

tools/vcli/bw/repl/repl.go
```diff
@@ -425,9 +425,10 @@ func runBQL(ctx context.Context, bql string, s storage.Store, chanSize, bulkSize
 		})
 		return nil, msg
 	}
+	nRows := res.NumRows()
```
rbkloss, line 428: I thing the old way is actually better, there is less variables and res.NumRows is pretty self explanatory, maybe res is not but that is another fix.

rbkloss (approved):

bql/planner/data_access.go
rogerlucena: The reason here to cache the `lo.String()` to a `string` variable is a bit different, I tried to explain it in the commit messages as well: the `lo` above is a **pointer** to a `storage.LookupOptions` struct, and this struct referenced by the pointer is changed multiple times in code (when setting global time bounds, when adding or removing `FilterOptions` to/from it, and so on). This way, depending on the time the closure is called in the tracer we may see a different `lo.String()` in the tracing output compared to the one we really wanted to trace at the moment of the `.Trace` call.

I found this problem when reviewing some traces and found an inconsistency with the `"Setting global lookup options to %s"` tracing message (it was showing a non-nil `FilterOptions` there, but at the moment of setting global lookup options we do not have any `FilterOptions` set yet!). This was a bit subtle to debug, and the way to correct it was just to pass `lo.String()` to a local variable before sending it to the tracer.

To avoid this problem happens again, I decided to repeat this process for all places in code on which the `lo` pointer was being captured by a tracing closure (commit `7a148a1`).

tools/vcli/bw/repl/repl.go
rogerlucena: What I wanted here was just to not capture the entire result table `res` inside the tracing closure just to have access to its `NumRows` later. The argument here is the same as the one for passing a `g.ID()` (graph ID) to a local variable `gID` before sending it to the tracer: we do not want to capture the entire graph inside the closure just to get its string `ID` later.

Issue #120, opened 2020-07-27
Create documentation about OPTIONAL keyword


thiagovas:
The OPTIONAL keyword was introduced in #89, but we're lacking documentation on it.

rogerlucena:
I believe this Issue can be closed as soon as the PR #161 is merged, since the `OPTIONAL` keyword is being documented there including some practical examples as well.

PR #161, opened 2020-11-17
Update and improve documentation

rogerlucena:
This PR comes to **improve** the current BadWolf documentation, making it more organized and clear, while also adding **updates** on the latest features added to BW (both to its Query Language - BQL and Command Line Interface - `bw` CLI).

Among the main topics documented with this PR we have:

- **`FILTER`** clause: how it works, why it is useful and how to use it (using the `latest`, `isImmutable` and `isTemporal` filter functions to exemplify);

- How to add support for a **new `FILTER` function** in the BadWolf code: instructions step by step were put into a separate file (`support_new_filter_function.md`), referenced by the documentation in `bql.md`;

- How to activate and use **`pprof` profiling** in BadWold (CLI);

- How to configure **verbosity** levels for **tracing** in BadWolf (CLI), and what each verbosity level means;

- Clarifications on the **graph pattern enforcement** inside `WHERE`;

- **`OPTIONAL`** keyword (as pointed by the Issue #120);

- More **complete `HAVING`** clause: support for comparisons involving `ID`, `TYPE`, `AT` and time anchor bindings, in addition to equalities for predicates and nodes too;

- Clarify that using `AFTER`, `BEFORE` and `BETWEEN` **does not automatically exclude immutable triples** from the query result: they are still shown there along with the temporal ones that are in the specified time interval. Propose the use of `FILTER isTemporal` if the user wants to see only temporal triples in the result (since the above specified behavior will be kept to do not break current users already relying on it);

- Be **clear** that the syntactic sugar of using a comma inside square brackets to reference time bounds, as in `"bought"@[,]`, makes sense only inside the `WHERE` scope;

- Show examples of `AT` being used for objects as well (in a reification scenario);

- Comment about the **trailing dot** being now **optional** for the last clause inside `WHERE` (as recommended by W3C);

- Add **more** practical **BQL examples** in `badwolf/examples/bql` (especially to cover `OPTIONAL`, `FILTER` and the more complete `HAVING` clause we currently support).


examples/bql/example_3.bql
```diff
@@ -44,5 +56,52 @@ WHERE {
...
+# What are the triples in the ?info graph having either "height_cm"@[]
+# or "born_in"@[] as predicate?
+SELECT ?subj, ?pred, ?obj
+FROM ?info
+WHERE {
+  ?subj ?pred ?obj
+}
+HAVING (?pred = "height_cm"@[]) OR (?pred = "born_in"@[]);
+
+# N.B. for nodes and predicates you are allowed to use equalities inside the
```
rbkloss, line 93: N.B stands for what?

examples/bql/example_4.bql
```diff
@@ -0,0 +1,78 @@
...
+SELECT ?s, ?p, ?o
+FROM ?info
+WHERE {
+  ?s ?p ?o .
+  FILTER latest(?p)
+};
+# (note how "latest" automatically skips the immutable triples).
+
+# What are the triples of the ?info graph for which the object is an
+# immutable predicate (reification)?
```
rbkloss, line 69: I think reification can be dropped or elaborated that it is just an example.

One can add an triples with predicates as object in cases that are not restricted to reification.

docs/support_new_filter_function.md
```diff
@@ -0,0 +1,45 @@
...
+The `FILTER` specifications always arrive to the driver level in the form of a `FilterOptions` struct
+inside `storage.LookupOptions`. The `Operation` represents the `FILTER` function being applied (eg: `IsTemporal`),
+while `Field` represents the position of the graph clause it shall be applied to (subject, predicate or object)
+and `Value` encapsulates the second argument of the `FILTER` function (not applicable for all `Operations` - some
+like `IsTemporal` do not use it while others like `GreaterThan` do, see [Issue 129](https://github.com/google/badwolf/issues/129)).
+
+Then, to implement a given `FILTER` behavior in the driver level all the user shall do is use the instructions for data
+retrieval received inside `FilterOptions` and proceed accordingly for the specific storage infrastructure they have at hand.
+
+**N.B.** When implementing the driver side of the `FILTER` functions be attentive to the order to process the information
```
rbkloss, line 32: "pay attention" instead of be attentive?

rbkloss (approved):

examples/bql/example_3.bql
rogerlucena: ["Nota bene"](https://pt.wikipedia.org/wiki/Nota_bene#:~:text=Nota%20bene%20%E2%80%93%20Wikip%C3%A9dia%2C%20a%20enciclop%C3%A9dia%20livre), in the sense of "pay attention to". If you want me to change it for something else, just lmk.

docs/support_new_filter_function.md
rogerlucena: Done, thank you Kloss.

examples/bql/example_4.bql
rogerlucena: Done, thank you Kloss. =]

docs/command_line_tool.md
```diff
@@ -253,12 +253,41 @@ export <graph_names_separated_by_commas> <file_path>  - dumps triples from graph
...
 quit                                                  - quits the console.
-bql> 
 ```
 
+### Tracing in BadWolf
+
+BadWolf has its own tracer implemented, that can be enabled/disabled with the `start` and `stop` commands detailed above.
+The current BadWolf tracer supports 3 levels of verbosity:
+
+- **`1`** for **minimum** verbosity: only the most crucial and basic messages to understand the query processing flow are sent to
```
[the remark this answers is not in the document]
rogerlucena: Just removed this part, thank you Kloss.

thiagovas (approved):

PR #166, opened 2022-03-10
Optimize slice allocation in Badwolf.

thiagovas:
There are some cases in the codebase where we know the size of slices beforehand.
Thus, we can pre-allocate them instead of relying on append to dynamically increase the slices.
This change decreases the number of times we call runtime.growslice, optimizing the code a bit, and also helps GC.

### DIFF
--- a/bql/planner/planner.go
+++ b/bql/planner/planner.go
@@ -288,10 +288,6 @@ func (p *queryPlan) Type() string {
 
 // newQueryPlan returns a new query plan ready to be executed.
 func newQueryPlan(ctx context.Context, store storage.Store, stm *semantic.Statement, chanSize int, w io.Writer) (*queryPlan, error) {
-	bs := []string{}
-	for _, b := range stm.Bindings() {
-		bs = append(bs, b)
-	}
 	t, err := table.New([]string{})
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


bql/planner/planner.go
```diff
@@ -288,18 +288,14 @@ func (p *queryPlan) Type() string {
...
-	}
 	t, err := table.New([]string{})
 	if err != nil {
 		return nil, err
 	}
 	return &queryPlan{
 		stm:       stm,
 		store:     store,
-		bndgs:     bs,
+		bndgs:     stm.Bindings(),
```
rbkloss, line 298: Before `bs` was a copy of stm.Bindings() which it no longer is, is it ok to change the behavior in this way?

Also throughout the code you are specifying capacity for the new slices and appending each element,
did you check the performance of doing one `append(copy, source...)`? (I am curious about the difference)

rbkloss (approved):

bql/planner/planner.go
thiagovas: stm.Bindings() already returns []string, so I guess there isn't much need to copy its result to bs here.

I didn't get to run benchmarks after these changes, but there are links explaining the otimization, like:
1) https://medium.com/vendasta/golang-the-time-complexity-of-append-2177dcfb6bad
2) https://www.programming-books.io/essential/go/optimize-by-pre-allocating-slice-ee6501b596c9446da8b82e4ed35454a6

PR #170, opened 2022-03-31
Store a pointer to slice in `sync.Pool`.

sidnei:
Storing the slice directly causes the value to be placed on the heap
which means an additional allocation.

See: https://staticcheck.io/docs/checks#SA6002

### DIFF
--- a/triple/triple.go
+++ b/triple/triple.go
@@ -263,14 +263,13 @@ func (t *Triple) UUID() uuid.UUID {
 	// value needs to be placed on the heap which causes an extra allocation.
 	// See: https://staticcheck.io/docs/checks#SA6002
 	bp := bufPool.Get().(*[]byte)
-	b := *bp
-	// Even though 'b' is dirty when fetched from the pool we do not
+	// Even though 'bp' is dirty when fetched from the pool we do not
 	// clear it because the contents are always completely overwritten.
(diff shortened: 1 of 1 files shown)


triple/triple.go
```diff
@@ -259,10 +259,14 @@ func (t *Triple) Reify() ([]*Triple, *node.Node, error) {
 // implemented as the SHA1 UUID of the concatenated UUIDs of the subject,
 // predicate, and object.
 func (t *Triple) UUID() uuid.UUID {
-	b := bufPool.Get().([]byte)
+	// We need to store a pointer to the value in sync.Pool otherwise the
+	// value needs to be placed on the heap which causes an extra allocation.
+	// See: https://staticcheck.io/docs/checks#SA6002
+	bp := bufPool.Get().(*[]byte)
+	b := *bp
```
thiagovas, line 266: sry, do we need to have this auxiliary variable here?

Can we have something like this, instead?
```
        copy((*bp)[:16], t.s.UUID())
	copy((*bp)[16:32], t.p.UUID())
	copy((*bp)[32:], t.o.UUID())
	return uuid.NewSHA1(uuid.NIL, (*bp))
```

thiagovas (changes requested):

thiagovas (approved):

